PR #3, opened 2025-10-22
[FLINK-38494] Add Chinese docs folder

davidradl:
Adding the Chinese docs folder. Populating with the english version, that can be translated when required.

### DIFF
--- a/docs/content.zh/docs/datastream/http.md
+++ b/docs/content.zh/docs/datastream/http.md
@@ -1,0 +1,179 @@
+---
+title: HTTP 
+weight: 3
+type: docs
+aliases:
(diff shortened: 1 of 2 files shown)


davidradl:
@ferenc-csaky are you ok to review and merge this one please. I will then implement [FLINK-38306](https://issues.apache.org/jira/browse/FLINK-38306) to knit in these docs into master Flink

ferenc-csaky (approved):
LGTM.

@davidradl AFAIK the base folder structure should be `docs/content/docs/connectors`. I see that the current docs not placed like that too, so it's fine to merge this, but that should be covered in that follow-up change.

PR #4, opened 2025-10-24
[hotfix] Move docs content to the correct location

davidradl:
As per the review comment in #3 , moving docs to the correct location


davidradl:
@ferenc-csaky thanks for your comment on the last pr - moving the docs to the right location

PR #8, opened 2025-12-01
[FLINK-38743] Change connector identifiers

davidradl:
changing the connector identifiers so they do not clash with the GetInData jar identifiers; so the 2 jar files can coexist in the same flink runtime.

### DIFF
--- a/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceFactory.java
+++ b/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceFactory.java
@@ -74,6 +74,8 @@ import static org.apache.flink.table.types.utils.DataTypeUtils.removeTimeAttribu
 /** Http Lookup Table Source Factory. */
 public class HttpLookupTableSourceFactory implements DynamicTableSourceFactory {
 
+    private static final String IDENTIFIER = "http";
+
     private static DataTypes.Field columnToField(Column column) {
         return FIELD(
                 column.getName(),
@@ -141,7 +143,7 @@ public class HttpLookupTableSourceFactory implements DynamicTableSourceFactory {
 
     @Override
     public String factoryIdentifier() {
-        return "http";
+        return IDENTIFIER;
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


davidradl:
@ferenc-csaky please could you have a look at this one - thanks?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/sink/HttpDynamicTableSinkFactory.java
```diff
@@ -38,7 +38,7 @@
 /** Factory for creating {@link HttpDynamicSink}. */
 public class HttpDynamicTableSinkFactory extends AsyncDynamicTableSinkFactory {
 
-    public static final String IDENTIFIER = "http-sink";
+    public static final String IDENTIFIER = "http-apache-sink";
```
ferenc-csaky: I don't like `http-apache-sink`. I have no problem with renaming the other ones, cause the changed names are generic too, but otherwise I'd prefer renaming it downstream. If it's not possible, then give it some general name that does not tie it to anything specifically, e.g. `http-async-sink`.

ferenc-csaky (changes requested):

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceFactory.java
```diff
@@ -141,7 +141,7 @@ protected void validateHttpLookupSourceOptions(ReadableConfig tableOptions)
 
     @Override
     public String factoryIdentifier() {
-        return "rest-lookup";
+        return "http";
```
ferenc-csaky: Lets add a class level constant `IDENTIFIER` for this as well, that's the common practice.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/sink/HttpDynamicTableSinkFactory.java
davidradl: @ferenc-csaky Good idea. I was struggling to find a good name.

ferenc-csaky (approved):

PR #12, opened 2025-12-18
[FLINK-38816] use Flink 2 compatible deserialise

davidradl:
The version of deserialise method that is used by this connector is deprecated at Flink 2, and a Runtime Exception is thrown. This change means that the Flink 2 compatible deserialise method is now used. 

This is a port of the GetInData PR [HTTP 204](https://github.com/getindata/flink-http-connector/pull/204); getindata requires 90% code coverage, the code has been implemented to be testable and meet this requirement then ported as is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .gitignore.swp
 .project
 .settings
+.metals
 .DS_Store
 /.java-version
 .eslintcache
--- a/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
+++ b/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
@@ -18,6 +18,7 @@
 package org.apache.flink.connector.http.table.lookup;
 
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.functions.util.ListCollector;
 import org.apache.flink.api.common.serialization.DeserializationSchema;
 import org.apache.flink.configuration.ReadableConfig;
 import org.apache.flink.connector.http.HttpLogger;
@@ -32,7 +33,6 @@ import org.apache.flink.connector.http.status.HttpResponseChecker;
 import org.apache.flink.connector.http.utils.HttpHeaderUtils;
 import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.functions.FunctionContext;
-import org.apache.flink.util.Collector;
 import org.apache.flink.util.ConfigurationException;
 import org.apache.flink.util.StringUtils;
 
@@ -327,36 +327,10 @@ public class JavaNetHttpPollingClient implements PollingClient {
         }
     }
 
-    @VisibleForTesting
-    List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
+    private List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
         List<RowData> result = new ArrayList<>();
-        responseBodyDecoder.deserialize(rawBytes, createRowDataCollector(result));
+        responseBodyDecoder.deserialize(rawBytes, new ListCollector(result));
         return result;
-    }
-
-    @VisibleForTesting
-    Collector<RowData> createRowDataCollector(List<RowData> result) {
-        return new RowDataCollector(result);
-    }
-
-    /** A simple collector implementation that adds RowData records to a list. */
-    @VisibleForTesting
-    static class RowDataCollector implements Collector<RowData> {
-        private final List<RowData> result;
-
-        RowDataCollector(List<RowData> result) {
-            this.result = result;
-        }
-
-        @Override
-        public void collect(RowData record) {
-            result.add(record);
-        }
-
-        @Override
-        public void close() {
-            // No-op - nothing to clean up
-        }
     }
 
     @VisibleForTesting
--- a/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
+++ b/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
@@ -32,6 +32,7 @@ import org.apache.flink.connector.http.status.HttpResponseChecker;
 import org.apache.flink.connector.http.utils.HttpHeaderUtils;
 import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.functions.FunctionContext;
+import org.apache.flink.util.Collector;
 import org.apache.flink.util.ConfigurationException;
 import org.apache.flink.util.StringUtils;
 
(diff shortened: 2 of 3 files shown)


davidradl:
@ferenc-csaky Please could you have a look at this one?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -326,22 +327,49 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
...
+    }
+
+    @VisibleForTesting
+    Collector<RowData> createRowDataCollector(List<RowData> result) {
+        return new RowDataCollector(result);
+    }
+
+    /** A simple collector implementation that adds RowData records to a list. */
+    @VisibleForTesting
+    static class RowDataCollector implements Collector<RowData> {
```
ferenc-csaky: There is a `ListCollector` which does exactly this, so we can get rid of this class.

ferenc-csaky:
I am not sure I understand this change. The `DeserializationSchema` interface is exactly the same in [Flink 1.20](https://github.com/apache/flink/blob/release-1.20/flink-core/src/main/java/org/apache/flink/api/common/serialization/DeserializationSchema.java) and the current [master](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/serialization/DeserializationSchema.java), so I do not see what this solves.

davidradl:
> I am not sure I understand this change. The `DeserializationSchema` interface is exactly the same in [Flink 1.20](https://github.com/apache/flink/blob/release-1.20/flink-core/src/main/java/org/apache/flink/api/common/serialization/DeserializationSchema.java) and the current [master](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/serialization/DeserializationSchema.java), so I do not see what this solves.

@ferenc-csaky It is the [https://github.com/apache/flink/blob/fec9336e7d99500aeb9097e441d8da0e6bde5943/flink-formats/flink-json/src/main/java/org/apache/flink/formats/json/AbstractJsonDeserializationSchema.java#L115](https://github.com/apache/flink/blob/fec9336e7d99500aeb9097e441d8da0e6bde5943/flink-formats/flink-json/src/main/java/org/apache/flink/formats/json/AbstractJsonDeserializationSchema.java#L115) that is deprecated - I should have mentioned this class by name for clarity. You can see that at version 2 the deprecated method does not continue to work, as you might expect, it throws an Exception.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -326,22 +327,49 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
...
+    List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
+        List<RowData> result = new ArrayList<>();
+        responseBodyDecoder.deserialize(rawBytes, createRowDataCollector(result));
+        return result;
+    }
+
+    @VisibleForTesting
+    Collector<RowData> createRowDataCollector(List<RowData> result) {
+        return new RowDataCollector(result);
+    }
```
ferenc-csaky: Any specific reason for `createRowDataCollector`? I do not see why not call the ctor directly in `deserializeSingleValue`:
```java
    List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
        List<RowData> result = new ArrayList<>();
        responseBodyDecoder.deserialize(rawBytes, new ListCollector(result));
        return result;
    }
```

ferenc-csaky (changes requested):

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -326,22 +327,49 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
         }
     }
 
-    private List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
-        return Optional.ofNullable(responseBodyDecoder.deserialize(rawBytes))
-                .map(Collections::singletonList)
-                .orElse(Collections.emptyList());
+    @VisibleForTesting
```
RocMarshal, line 329: could the line be deleted?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -326,22 +327,49 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
...
+        return result;
+    }
+
+    @VisibleForTesting
+    Collector<RowData> createRowDataCollector(List<RowData> result) {
+        return new RowDataCollector(result);
+    }
+
+    /** A simple collector implementation that adds RowData records to a list. */
+    @VisibleForTesting
```
RocMarshal: ditto.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
davidradl: With the other change I agree - I will remove this tag and make the method private

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
davidradl: this class will be removed with the simplication Ferenc suggested.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -326,22 +327,49 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
...
+        }
+
+        @Override
+        public void close() {
+            // No-op - nothing to clean up
+        }
     }
 
-    private List<RowData> deserializeArray(byte[] rawBytes) throws IOException {
+    @VisibleForTesting
```
davidradl, line 354: I think we should keep this  @VisibleForTesting as it justifies why this method is not private.

davidradl:
@RocMarshal @ferenc-csaky thanks for your feedback - it is ready for review again.

davidradl:
changing to draft while I check something out

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -327,21 +328,22 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
     }
 
     private List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
-        return Optional.ofNullable(responseBodyDecoder.deserialize(rawBytes))
-                .map(Collections::singletonList)
-                .orElse(Collections.emptyList());
+        List<RowData> result = new ArrayList<>();
+        responseBodyDecoder.deserialize(rawBytes, new ListCollector(result));
+        return result;
```
ferenc-csaky: One thing I missed before, let's wrap it to `Collections.unmodifiableList(result)`, cause the prev. code returned an immutable list.

ferenc-csaky (commented):
Added 1 comment, other than that LGTM

.gitignore
```diff
@@ -5,6 +5,7 @@
 .gitignore.swp
 .project
 .settings
+.metals
```
davidradl, line 8: my IDE added this so put in an ignore for it.

ferenc-csaky (approved):

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -326,22 +326,42 @@ private Collection<RowData> deserialize(String responseBody) throws IOException
...
-                .orElse(Collections.emptyList());
+    @VisibleForTesting
+    List<RowData> deserializeSingleValue(byte[] rawBytes) throws IOException {
+        List<RowData> result = new ArrayList<>();
+        responseBodyDecoder.deserialize(rawBytes, new ListCollector(result));
+        return Collections.unmodifiableList(result);
     }
 
-    private List<RowData> deserializeArray(byte[] rawBytes) throws IOException {
+    private static class ListCollector implements org.apache.flink.util.Collector<RowData> {
```
ferenc-csaky, line 336: I just merged it and spotted this came back. Why?

davidradl: @ferenc-csaky My mistake - I meant this to be left unchanged - let me raise another PR to address this.

PR #13, opened 2026-01-06
[Flink-38864] Add ignore status completion state

davidradl:
Add ignore status completion state nd associated unit tests.


docs/content.zh/docs/connectors/table/http.md
```diff
@@ -503,6 +503,10 @@ this means that these columns will be null for nullable columns and hold a defau
 
 When using `http.source.lookup.continue-on-error` as true, consider adding extra metadata columns that will surface information about failures into your stream.
 
+Note that if Metadata columns are specified and the status code is ignored, then a row containing metadata columns will be produced. If
```
ferenc-csaky: nit: Metadata -> metadata

same in the other doc file

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -520,10 +524,14 @@ Metadata columns can be specified and hold http information. They are optional r
 | HTTP_ERROR_STATUS              | HTTP error status code              |
 | EXCEPTION                      | An Exception occurred               |
 | UNABLE_TO_DESERIALIZE_RESPONSE | Unable to deserialize HTTP response |
+| IGNORE_STATUS_CODE             | Status code was ignored             |
```
ferenc-csaky: nit: maybe "Status code is ignored"?

if change, same in the other doc file

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpTableLookupFunction.java
```diff
@@ -104,9 +104,6 @@ public Collection<RowData> lookup(RowData keyRow) {
         int physicalArity = -1;
 
         GenericRowData producedRow = null;
-        if (httpRowDataWrapper.shouldIgnore()) {
-            return Collections.emptyList();
```
ferenc-csaky, line 108: Why is this not necessary anymore?

ferenc-csaky (changes requested):

ferenc-csaky:
Also, currently this PR includes #12 as well, so we should not merge this in its current state.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpTableLookupFunction.java
davidradl: If there are metadata columns then we want to populate them appropriately for the ignore case

davidradl:
> Also, currently this PR includes #12 as well, so we should not merge this in its current state.

Good spot - I will remove  #12 from this PR

davidradl:
@ferenc-csaky It is ready for review again -thanks

ferenc-csaky (approved):

PR #17, opened 2026-02-06
FLINK-39040 Additional body json for GenericJsonAndUrlQueryCreator

davidradl:


davidradl:
@ferenc-csaky are you OK to have a look at this please.

qinf:
Hi @davidradl seems this PR linked a wrong Flink issue. Maybe [FLINK-39040](https://issues.apache.org/jira/projects/FLINK/issues/FLINK-39040) is the wright one.

davidradl:
> Hi @davidradl seems this PR linked a wrong Flink issue. Maybe [FLINK-39040](https://issues.apache.org/jira/projects/FLINK/issues/FLINK-39040) is the wright one.

Thanks for the feedback - I corrected the PR title on Friday. The branch name is still incorrect - but that should not matter - as it does not effect the code.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -145,9 +149,36 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
...
+                        if (additionalNode.isObject()) {
+                            ObjectNode additionalObjectNode = (ObjectNode) additionalNode;
+                            // Merge all fields from additional JSON into the body
+                            // This preserves nested objects and arrays as-is
+                            additionalObjectNode
+                                    .fields()
+                                    .forEachRemaining(
+                                            entry ->
+                                                    bodyJsonObject.set(
+                                                            entry.getKey(), entry.getValue()));
```
ferenc-csaky: I am wondering what would be the desired behavior here? If i have a key in `additionalObjectNode` that `bodyJsonObject` also has, the current logic will overwrite the original value. Do we wanna do that? By default I as a user may not want that. Maybe I want to add some default values to my data, but I definitely do not want to overwrite the field if it exists already.

I am thinking that we may also make this configurable and introduce a flag to control this behavior. So to make them as a group we could have something like:
- `http.request.additional-body-json.content` (just renaming the current config option)
- `http.request.additional-body-json.overwrite` (boolean config flag to ocntrol overwrite behavior)

WDYT?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -92,6 +92,18 @@ public class GenericJsonAndUrlQueryCreatorFactory implements LookupQueryCreatorF
                                     + "The expected format of the map is:"
                                     + "<br>"
                                     + " key1:value1,key2:value2");
+    public static final ConfigOption<String> ADDITIONAL_REQUEST_JSON =
```
ferenc-csaky: nit: I'd name this constant `REQUEST_ADDITONAL_BODY_JSON` to match the already existing naming pattern.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
davidradl: hm - good point. I am thinking that on the whole we would not want to overwrite an existing key - because it is a lookup key - that is supposed to change on a per request basis so it targets different enrichment content per request. I propose I validate that the keys in the additional json to ensure they do not match join keys. 
WDYT?

ferenc-csaky: You are also right, I ignored the wider context here, +1 for your suggestion.

davidradl:
@ferenc-csaky ready for review again!

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -152,6 +176,62 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+    private void validateAdditionalJsonDoesNotOverrideJoinKeys(
+            List<String> requestBodyFields, String additionalRequestJson) {
+        if (additionalRequestJson == null || additionalRequestJson.trim().isEmpty()) {
+            return;
+        }
+
+        try {
+            // Parse the additional JSON to get field names
+            ObjectMapper mapper = new ObjectMapper();
+            JsonNode jsonNode = mapper.readTree(additionalRequestJson);
```
seb-pereira: The value of `http.request.additional-body-json` seems to be parsed twice:
- here by `mapper.readTree(additionalRequestJson`
- at merge time in [GenericJsonAndUrlQueryCreator](https://github.com/apache/flink-connector-http/pull/17/changes#diff-e73bc96c08ecfe30b58370e4631ac5f6c936ac60dd831475508a9995b8f01f76R158)

`additionalRequestJson` could be parsed once here and `validateAdditionalJsonDoesNotOverrideJoinKeys` to return a JsonNode that would be used in [createLookupQuery](https://github.com/apache/flink-connector-http/pull/17/changes#diff-e73bc96c08ecfe30b58370e4631ac5f6c936ac60dd831475508a9995b8f01f76L114) for the merge ([example in draft PR](https://github.com/davidradl/flink-connector-http/pull/1/changes)). WDYT?

davidradl: @seb-pereira I like it. Implementing

davidradl:
@seb-pereira ready for review again!

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -152,6 +178,67 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+     */
+    private ObjectNode createAdditionalObjectNode(
+            List<String> requestBodyFields, String additionalRequestJson) {
+        if (additionalRequestJson == null || additionalRequestJson.trim().isEmpty()) {
+            return null;
+        }
+
+        try {
+            // Parse the additional JSON once to avoid re-parsing on every lookup
+            ObjectMapper mapper = new ObjectMapper();
```
seb-pereira: Minor: for consistency with https://github.com/apache/flink-connector-http/pull/17/changes#diff-e73bc96c08ecfe30b58370e4631ac5f6c936ac60dd831475508a9995b8f01f76R165

```suggestion
            ObjectMapper mapper = ObjectMapperAdapter.instance();
```

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -152,6 +178,67 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+            }
+
+            // If there are conflicts, throw exception with all conflicting fields
+            if (!conflictingFields.isEmpty()) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "The http.request.additional-body-json option should not override join keys, "
+                                        + "as join keys are expected to target different enrichments on a request basis. "
+                                        + "Found conflicting field%s: %s",
+                                conflictingFields.size() > 1 ? "s" : "",
```
ferenc-csaky: I'd just spare this ternary if that makes this pretty hard to read and do `field(s): %s`

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -152,6 +178,67 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+        }
+
+        try {
+            // Parse the additional JSON once to avoid re-parsing on every lookup
+            ObjectMapper mapper = ObjectMapperAdapter.instance();
+            JsonNode jsonNode = mapper.readTree(additionalRequestJson);
+
+            if (!jsonNode.isObject()) {
+                throw new IllegalArgumentException(
+                        "The http.request.additional-body-json must be a valid JSON object.");
```
ferenc-csaky: I'd rather refer the key as `REQUEST_ADDITIONAL_BODY_JSON.key()` cause if 1 year from now that key changes for some reason hardcoding this may won't be detected. Same for the rest of the error messages

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -152,6 +178,67 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+
+            // Collect all conflicting fields
+            Set<String> conflictingFields = new HashSet<>();
+            Iterator<String> fieldNames = jsonNode.fieldNames();
+            while (fieldNames.hasNext()) {
+                String fieldName = fieldNames.next();
+                if (joinKeyNames.contains(fieldName)) {
+                    conflictingFields.add(fieldName);
+                }
+            }
```
ferenc-csaky: This can be simplified:
```java
Set<String> conflictingFields = new HashSet<>();
jsonNode.fieldNames().forEachRemaining(conflictingFields::add);
conflictingFields.retainAll(joinKeyNames);
```

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -152,6 +178,67 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+     * Creates and validates the additional JSON node from configuration. This method parses the
+     * JSON once during factory creation to avoid re-parsing on every lookup request, improving
+     * runtime performance.
+     *
+     * @param requestBodyFields the list of request body field names (join keys)
+     * @param additionalRequestJson the additional JSON string to validate and parse
+     * @return the parsed ObjectNode, or null if no additional JSON is provided
+     * @throws IllegalArgumentException if the JSON is invalid or contains conflicting fields
+     */
+    private ObjectNode createAdditionalObjectNode(
```
ferenc-csaky: Maybe `getValidatedAdditionalObjectNode`? Since we emphasize validated, I think it's better to reflect that in the method name, then we don't need that comment when we call this method at L130.

ferenc-csaky (changes requested):

davidradl:
@ferenc-csaky addressed the review feedback - the while loop simplification was nice :-).

ferenc-csaky (approved):
LGTM

PR #20, opened 2026-02-13
[FLINK-39087] Upgrade wiremock

davidradl:


davidradl:
@ferenc-csaky could you have a look at this one please?

flink-connector-http/pom.xml
```diff
@@ -306,10 +306,61 @@ under the License.
         </dependency>
 
         <dependency>
-            <groupId>com.github.tomakehurst</groupId>
+            <groupId>org.wiremock</groupId>
             <artifactId>wiremock</artifactId>
             <version>${wiremock.version}</version>
             <scope>test</scope>
+            <exclusions>
```
ferenc-csaky: I guess you needed to replicate exclusions here, cause this definition is not inheriting the one defined in the root POM, which I think is wrong. You also changed the `groupId` here, but not in the root POM. 

Replicating the exclusions here is  unnecessary, we just need to defined this de to inherit the one in the root POM, so I'd match `groupId` in the root POM, and remove the explicit `version` and all the `exclusions` here.

flink-connector-http/pom.xml
```diff
@@ -306,10 +306,61 @@ under the License.
...
+                    <groupId>com.networknt</groupId>
+                    <artifactId>json-schema-validator</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>net.javacrumbs.json-unit</groupId>
+            <artifactId>json-unit-core</artifactId>
+            <version>2.40.1</version>
```
ferenc-csaky, line 317: I see that this is not enforced in this project, so it should not be the scope of this PR, but in a subsequent change I'd centralize all dep and plugin versions in the root POM props in alphabetic order, cause it's a lot more human friendly.

And the same i true for exclusions, cause if we start excluding per-module it will be a lot more effort top maintain convergence cause it will be possible that the same dep is included with a different transitive version set.

ferenc-csaky (changes requested):

flink-connector-http/pom.xml
davidradl: Yes my mistake. I have changed it

flink-connector-http/pom.xml
davidradl: Agreed - I will do this under [FLINK-39101](https://issues.apache.org/jira/browse/FLINK-39101)

ferenc-csaky (approved):

PR #21, opened 2026-02-16
[FLINK-39101] Rationalize POM file versions and exclusions

davidradl:


davidradl:
@ferenc-csaky please could you take a look at this one please - this is the implementation of your suggested change from [PR 20](https://github.com/apache/flink-connector-http/pull/20)

pom.xml
```diff
@@ -64,44 +64,51 @@ under the License.
              TODO is this still valid in Flink connector.
              -->
 
-        <flink.version>1.20.0</flink.version>
```
ferenc-csaky, line 67: This may be opinionated, but I'd define some groups for the props and separate them by empty lines, or even add comment headers for them. Maybe something like:

```xml
<properties>
  <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
  <target.java.version>11</target.java.version>
  <scala.binary.version>2.12</scala.binary.version>
  <maven.compiler.source>${target.java.version}</maven.compiler.source>
  <maven.compiler.target>${target.java.version}</maven.compiler.target>

  <!-- Flink versions-->
  ...

  <!-- 3rd party versions-->
  ...

  <!-- Plugin versions-->
  ...

  <!-- Config args-->
  <flink.surefire.baseArgLine>-XX:+UseG1GC -Xms256m -XX:+IgnoreUnrecognizedVMOptions
    ${flink.connector.module.config}
  </flink.surefire.baseArgLine>
  <!-- can be removed with maven-spotless-plugin:2.38+ -->
  <spotless.skip>false</spotless.skip>
</properties>
```

Feel free to shape this in any way you think better, I just tried to come up with something that would be useful to me and I tend to follow in way it makes sense in the given project.

ferenc-csaky (approved):

PR #23, opened 2026-02-20
[hotfix] change parent to 1.1

davidradl:
The idea is that we will update main to be Flink version 1 compatible , create the version 1 branch then upgrade main to be compiled against Flink version 2 and to use the version 2 parent.


davidradl:
hi @ferenc-csaky , please could you process this, thank you

ferenc-csaky (approved):

PR #28, opened 2026-03-02
[hotfix] Fix grammar

mariyageorge01:
Fix grammar and avoid ambiguity in documentation


docs/content.zh/docs/connectors/datastream/http.md
```diff
@@ -72,15 +72,15 @@ These options are specified on the builder using the setProperty method.
...
+| flink.connector.http.sink.writer.request.mode             | optional | Sets the Http Sink request submission mode. Two modes are available: `single` and `batch`. Defaults to `batch` if not specified. |
 | flink.connector.http.sink.request.batch.size              | optional | Applicable only for `flink.connector.http.sink.writer.request.mode = batch`. Sets number of individual events/requests that will be submitted as one HTTP request by HTTP sink. The default value is 500 which is same as HTTP Sink `maxBatchSize` |
 
 
 
 ### Batch submission mode
 
-By default, batch size is set to 500 which is the same as Http Sink's `maxBatchSize` property and has value of 500.
-The `maxBatchSize` property sets maximal number of events that will be buffered by Flink runtime before passing it to Http Sink for processing.
+By default, batch size is set to 500 which is the same as Http Sink's `maxBatchSize` property.
```
davidradl: I don't understand this sentence before! when we say batch size - I assume there is a batch size property, if so we should refer to it by property name. The way it reads is ambiguous as we day the batch size is 500 bit the same as a configurable property ? I assume we are saying that the default values for both these batch size configuration properties is 500, but the way it is phrased is not clear.

mariyageorge01: I added the property name.

docs/content.zh/docs/connectors/datastream/http.md
```diff
@@ -72,15 +72,15 @@ These options are specified on the builder using the setProperty method.
...
 | flink.connector.http.sink.request.batch.size              | optional | Applicable only for `flink.connector.http.sink.writer.request.mode = batch`. Sets number of individual events/requests that will be submitted as one HTTP request by HTTP sink. The default value is 500 which is same as HTTP Sink `maxBatchSize` |
 
 
 
 ### Batch submission mode
 
-By default, batch size is set to 500 which is the same as Http Sink's `maxBatchSize` property and has value of 500.
-The `maxBatchSize` property sets maximal number of events that will be buffered by Flink runtime before passing it to Http Sink for processing.
+By default, the `flink.connector.http.sink.request.batch.size` property is set to 500, which matches the default value of Http Sink's `maxBatchSize` property.
+The `maxBatchSize` property sets the maximum number of events that will be buffered by Flink runtime before passing it to Http Sink for processing.
```
davidradl: I was asking myself :what do these mean.  It looks like the connector buffers records internally up to size maxBatchSize, this will be flushed when it reaches maxBatchSize or sink.flush-buffer.timeout is met.

Example without timeout:
maxBatchSize = 500
batch.size = 100 
What happens:
Flink buffers up to 500 records (maxBatchSize)
When flush is triggered due to reaching 500, those 500 records are split into 5 HTTP requests

So this mechanism decouples the rate and size of records coming in vs how we send them as http requests.

We should try to say this in the text.

mariyageorge01: Updated the doc with more details.

docs/content.zh/docs/connectors/datastream/http.md
```diff
@@ -92,8 +98,10 @@ HttpSink.<String>builder()
       .build();
 ```
 
-### Single submission mode
-In this mode every processed event is submitted as individual HTTP POST/PUT request.
+In this example, the default `sink.batch.max-size` of 500 is used. When Flink flushes 500 records, the HTTP Sink splits them into 10 HTTP POST requests (50 records each).
```
davidradl, line 101: I think the single submission mode, is does as the original text said. The `PerRequestSubmitter` class does not look at the batching configuration. This is all done in the `BatchRequestSubmitter ` class. 

I suggest we add what the config parameter is that means we are in Single submission mode

mariyageorge01: I haven't changed the content for single submission mode. 

I added a common header for Batch submission mode and Single submission mode (like in Table API doc) and formatted the headings for better clarity.

PR #29, opened 2026-03-07
[FLINK-39222] Add licenses to files and correct Flink version

davidradl:


pom.xml
```diff
@@ -34,7 +34,7 @@ under the License.
 
     <artifactId>flink-connector-http-parent</artifactId>
     <name>Flink : Connectors : http : Parent</name>
-    <version>1.0-SNAPSHOT</version>
+    <version>1.0.0</version>
```
ferenc-csaky: We should keep `SNAPSHOT` versions on all branches, and only set the release version with the releasing commit as part of the release process.

ferenc-csaky (changes requested):

ferenc-csaky (approved):

PR #32, opened 2026-03-12
[FLINK-39365][connector-http] Support custom HTTP headers via ConfigOption

featzhang:
Add formal ConfigOption definitions for custom HTTP headers in both the HTTP lookup source and HTTP dynamic sink connectors.

Previously, custom HTTP headers were only configurable via raw properties using the prefixes 'http.source.lookup.header.' and 'http.sink.header.', but these were not declared as official ConfigOptions. This meant no IDE autocomplete support, missing documentation, and unclear user guidance.

Changes:
- Add SOURCE_LOOKUP_HEADERS ConfigOption (Map type) to HttpLookupConnectorOptions with prefix 'http.source.lookup.header'
- Add SINK_HEADERS ConfigOption (Map type) to HttpDynamicSinkConnectorOptions with prefix 'http.sink.header'
- Register both options in the respective factory's optionalOptions() so Flink's table option validator recognizes them
- Add tests verifying custom headers can be specified in SQL DDL without triggering ValidationException

The header values are still passed through Properties and read by the existing HttpHeaderUtils.prepareHeaderMap() mechanism, so the runtime behavior is unchanged. The new ConfigOptions serve primarily as documentation and validation anchors.


docs/content.zh/docs/connectors/datastream/http.md
```diff
@@ -52,7 +52,7 @@ These options are specified on the builder using the setProperty method.
...
 |---------------------------------------------------------|----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| connector                                               | required | Specify what connector to use. For HTTP Sink it should be set to _'http-sink'_.                                                                                                                                                                  |
+| connector                                               | required | Specify what connector to use. For HTTP Sink it should be set to _'http-async-sink'_.                                                                                                                                                                  |
```
davidradl, line 55: can you give me context about this change please. I assume both modes are supported.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupConnectorOptions.java
```diff
@@ -227,4 +229,24 @@ public class HttpLookupConnectorOptions {
...
+     *
+     * <p>Headers are specified as key-value pairs using the prefix {@code
+     * http.source.lookup.header.}, for example:
+     *
+     * <pre>{@code
+     * 'http.source.lookup.header.Content-Type' = 'application/json',
+     * 'http.source.lookup.header.Authorization' = 'Bearer my-token'
+     * }</pre>
+     */
+    public static final ConfigOption<Map<String, String>> SOURCE_LOOKUP_HEADERS =
```
davidradl, line 244: we can already support custom headers - see [https://github.com/apache/flink-connector-http/blob/main/docs/content/docs/connectors/table/http.md#http-headers](https://github.com/apache/flink-connector-http/blob/main/docs/content/docs/connectors/table/http.md#http-headers)

docs/content.zh/docs/connectors/datastream/http.md
featzhang: Thanks for the question, @davidradl.

This is a documentation fix I noticed while working on this PR. The actual connector identifier registered in `HttpDynamicTableSinkFactory` is http-async-sink, not http-sink. The original doc had an incorrect value that could confuse users when defining their DDL.

https://github.com/apache/flink-connector-http/blob/21b3d9a124fa8294f85c1ae0935ea9fa8bae62f7/flink-connector-http/src/main/java/org/apache/flink/connector/http/table/sink/HttpDynamicTableSinkFactory.java#L41

I included it in this PR since it's a small doc correction, but I'm happy to split it into a separate commit or even a separate PR if you prefer to keep things more focused.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupConnectorOptions.java
featzhang: Thanks @davidradl, you're right that custom headers are already functionally supported via the http.source.lookup.header. and http.sink.header. property prefixes.

This PR does not change the runtime behavior at all — headers are still read by the existing HttpHeaderUtils.prepareHeaderMap() mechanism. The goal here is to formalize these existing properties as official ConfigOption declarations. Without this change, there are a few practical issues:

Validation errors: When users specify headers like 'http.source.lookup.header.Content-Type' = 'application/json' in SQL DDL, Flink's table option validator throws a ValidationException because these keys are not registered as recognized options in optionalOptions().

Discoverability: Formal ConfigOption definitions enable IDE autocomplete and appear in generated documentation, making it easier for users to find and use these settings.

Consistency: Other Flink connectors (e.g., Kafka, JDBC) declare all their configuration as ConfigOption — this aligns the HTTP connector with that standard practice.

In short, this is about promoting existing "hidden" configuration to first-class ConfigOption status for better validation, documentation, and user experience.

docs/content.zh/docs/connectors/datastream/http.md
davidradl: that is fine - this is likely because we want all the names of things to be unique compared to getindata so they can coexist.

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -472,6 +477,30 @@ another format name.
...
+```roomsql
+CREATE TABLE http_sink (
+  id bigint,
+  some_field string
+) WITH (
+  'connector' = 'http-async-sink',
+  'url' = 'http://example.com/myendpoint',
+  'format' = 'json',
+  'http.sink.header.Content-Type' = 'application/json',
+  'http.sink.header.Authorization' = 'Bearer my-token',
```
davidradl, line 500: fyi there is OIDC support for lookup joins  - which uses bearer tokens under the covers. This is another candidate to be implemented for sinks.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupConnectorOptions.java
davidradl: @featzhang that makes sense- we just need to be careful about precedence and to give recommended ways of configuring. i.e. recommend the new option for the reasons you say

PR #34, opened 2026-03-15
[FLINK-39368][connector-http] Add User-Agent configuration support for HTTP sink and lookup

featzhang:

Add a dedicated `http.user.agent` option so that requests issued by the
HTTP connector carry an identifiable `User-Agent` header out of the box,
for both the lookup source and the async sink.

Without this option, the underlying Java `HttpClient` sends requests
without any `User-Agent` header, which makes server-side logging,
rate limiting and access auditing harder. Users could already override
the header via `http.source.lookup.header.User-Agent` /
`http.sink.header.User-Agent`, but there was no sensible default.

## Brief change log

- Add `http.user.agent` ConfigOption to `HttpLookupConnectorOptions`
  and `HttpDynamicSinkConnectorOptions`. Default value is
  `flink-connector-http` (matches the connector artifact identifier).
- Apply the option in `RequestFactoryBase` (lookup) and
  `JavaNetSinkHttpClient` (sink).
- Reject setting `User-Agent` simultaneously via `http.user.agent` and
  via the corresponding `http.*.header.User-Agent` entry with a clear
  `IllegalArgumentException`.
- Register both `SINK_HEADERS` / `SOURCE_LOOKUP_HEADERS` (from
  FLINK-HTTP-3) and the new `USER_AGENT` option in the factory
  `optionalOptions()` lists.
- Document the new option in the lookup and sink options tables and
  add a short `User-Agent` subsection with an example DDL.
- Tests cover default value, custom value and the conflict path.

## Verifying this change

- `JavaNetSinkHttpClientTest#shouldBuildClientWithoutHeaders` now
  asserts that the default `User-Agent: flink-connector-http` is
  attached when no header configuration is given.
- `JavaNetSinkHttpClientTest#shouldBuildClientWithCustomUserAgent` and
  `JavaNetHttpPollingClientTest#shouldBuildClientWithCustomUserAgent`
  assert that a user-provided value is honoured.
- `JavaNetSinkHttpClientTest#shouldThrowErrorWhenUserAgentConflict`
  and the lookup counterpart assert that the conflict detection fires
  with a helpful message.

- Dependencies: no
- The public API: yes — new `http.user.agent` ConfigOption
- The runtime per-record code paths: only by attaching one extra
  header to outgoing requests
- Anything that affects deployment or recovery: no

## Documentation

- Does this pull request introduce a new feature: yes
- If yes, how is the feature documented: in
  `docs/content/docs/connectors/table/http.md` and the `content.zh`
  equivalent, under the lookup options table, sink options table and
  a dedicated `User-Agent` subsection

### DIFF
--- a/docs/content.zh/docs/connectors/table/http.md
+++ b/docs/content.zh/docs/connectors/table/http.md
@@ -51,6 +51,7 @@ The HTTP source connector supports [Lookup Joins](https://nightlies.apache.org/f
       * [For HTTP responses](#for-http-responses)
     * [Default Query Creator Implementation](#default-query-creator-implementation)
     * [Http headers](#http-headers)
+    * [User-Agent](#user-agent)
     * [Timeouts](#timeouts)
     * [Source table HTTP status code](#source-table-http-status-code)
     * [Retries and handling errors (Lookup source)](#retries-and-handling-errors-lookup-source)
@@ -208,6 +209,7 @@ Note the options with the prefix _http_ are the HTTP connector specific options,
 | http.request.query-param-fields-with-key                               | optional | A map of column names to query parameter keys. See the [Query Parameter Mapping](#query-parameter-mapping) section for details and examples.                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
 | http.request.body-template                                             | optional | Used for the `http-generic-json-url` query creator. A JSON template string for constructing the request body for PUT and POST operations. Use `{{fieldName}}` placeholders to reference top-level columns from the lookup table. Supports creating complex nested JSON structures with both placeholders and literal values. See the [Body Template](#body-template) section for details and examples.                                                                                                                                                                                                     |
 | http.request.url-map                                                   | optional | Used for the `http-generic-json-url` query creator. The map of insert names to column names used as url segments. Parses a string as a map of strings. For example if there are table columns called `customerId` and `orderId`, then specifying value `customerId:cid,orderID:oid` and a url of https://myendpoint/customers/{cid}/orders/{oid} will mean that the url used for the lookup query will dynamically pickup the values for `customerId`, `orderId` and use them in the url e.g. https://myendpoint/customers/cid1/orders/oid1. The expected format of the map is: `key1:value1,key2:value2`. |
+| http.user.agent                                                        | optional | The value of the `User-Agent` HTTP header sent with every lookup request. Defaults to `flink-connector-http`. Setting this option together with `http.source.lookup.header.User-Agent` is not allowed and will fail with a configuration error.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
 
 ### Query Creators
 
@@ -448,6 +450,32 @@ CREATE TABLE http-lookup (
 Note that when using OIDC, it adds an `Authentication` header with the bearer token; this will override
 an existing `Authorization` header specified in configuration.
 
+### User-Agent
+
+The `User-Agent` header value can be configured via the dedicated `http.user.agent` option. When not
+specified, the connector sends every request with `User-Agent: flink-connector-http`. This makes the
+connector identifiable to the target server out of the box and is useful for server-side logging,
+rate limiting and access auditing.
+
+```roomsql
+CREATE TABLE Customers (
+  id INT,
+  name STRING,
+  email STRING
+) WITH (
+  'connector' = 'http',
+  'url' = 'http://api.example.com/customers',
+  'format' = 'json',
+  'http.user.agent' = 'MyApp/1.0'
+)
+```
+
+The same option is also available for the HTTP sink, see the [Sink Connector Options](#sink-connector-options) table.
+
+Setting `User-Agent` through both `http.user.agent` and `http.source.lookup.header.User-Agent`
+(or `http.sink.header.User-Agent` for the sink) is ambiguous and is rejected at table creation time
+with an `IllegalArgumentException`. Configure it through exactly one of the two options.
+
 ### Timeouts
 Lookup Source is guarded by two timeout timers. First one is specified by Flink's AsyncIO operator that executes `AsyncTableFunction`.
 The default value of this timer is set to 3 minutes and can be changed via `table.exec.async-lookup.timeout` [option](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/config/#table-exec-async-lookup-timeout).
@@ -547,6 +575,7 @@ another format name.
 | http.sink.writer.thread-pool.size         | optional | Sets the size of pool thread for HTTP Sink request processing. Increasing this value would mean that more concurrent requests can be processed in the same time. If not specified, the default value of 1 thread will be used.     |
 | http.sink.writer.request.mode             | optional | Sets the Http Sink request submission mode. Two modes are available: `single` and `batch`. Defaults to `batch` if not specified. |
 | http.sink.request.batch.size              | optional | Applicable only for `http.sink.writer.request.mode = batch`. Sets number of individual events/requests that will be submitted as one HTTP request by HTTP sink. The default value is 500 which is same as HTTP Sink `maxBatchSize` |
+| http.user.agent                           | optional | The value of the `User-Agent` HTTP header sent with every sink request. Defaults to `flink-connector-http`. Setting this option together with `http.sink.header.User-Agent` is not allowed and will fail with a configuration error.                                                                                                |
 
 ### Sink table HTTP status codes
 You can configure a list of HTTP status codes that should be treated as errors for HTTP sink table.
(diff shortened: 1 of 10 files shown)


flink-connector-http/src/main/java/org/apache/flink/connector/http/sink/httpclient/JavaNetSinkHttpClient.java
```diff
@@ -73,6 +73,12 @@ public JavaNetSinkHttpClient(
                         properties,
                         headerPreprocessor);
 
+        // Add User-Agent header if configured
```
davidradl: Please raise a Jira to describe what you are look to do here ; with some examples. It is possible to add your own custom headers at the moment - why should this header have its own config.

featzhang: You're right that the existing prefix mechanism already works:

  'http.source.lookup.header.User-Agent' = 'my-agent/1.0'
  'http.sink.header.User-Agent'          = 'my-agent/1.0'

  The main motivation for a dedicated option is the default value — without any configuration today, the connector sends requests with no User-Agent header
  at all. Many HTTP servers use User-Agent for logging and traffic attribution, so having the connector identify itself as flink-http-connector out of the
  box seems useful without requiring every user to set it explicitly.
  A secondary benefit is discoverability: a named ConfigOption appears in documentation and tooling, whereas the prefix-based keys are not enumerated anywhere.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/sink/HttpDynamicSinkConnectorOptions.java
```diff
@@ -41,4 +41,12 @@ public class HttpDynamicSinkConnectorOptions {
...
                     .stringType()
                     .defaultValue(Slf4jHttpPostRequestCallbackFactory.IDENTIFIER);
+
+    public static final ConfigOption<String> USER_AGENT =
+            ConfigOptions.key("http.user.agent")
+                    .stringType()
+                    .defaultValue("flink-http-connector")
+                    .withDescription(
+                            "The User-Agent header value for HTTP sink requests. "
+                                    + "Default value is 'flink-http-connector'.");
```
davidradl: maybe default to `flink-connector-http`. `flink-http-connector i`s the getindata identifier for the HTTP connector we moved from.

flink-connector-http/src/test/java/org/apache/flink/connector/http/sink/httpclient/JavaNetSinkHttpClientTest.java
```diff
@@ -134,5 +139,50 @@ public void shouldBuildClientWithHeaders(RequestSubmitterFactory requestSubmitte
                 "Cache-Control",
                 "no-cache, no-store, max-age=0, must-revalidate");
         assertPropertyArray(headersAndValues, "Access-Control-Allow-Origin", "*");
+        // Default User-Agent header should also be present
+        assertPropertyArray(headersAndValues, "User-Agent", "flink-http-connector");
```
davidradl: it looks like from the code that if a user agent header was added in the headers part then that would take precedence over the explicit user agent setting. Is this what we want? I suggest a config error if we have a clash like this

davidradl (changes requested):
please add documentation

featzhang:
Hi @davidradl,

I've added comprehensive documentation for the `http.user.agent` configuration option in commit dddfa5b.

**What's Added:**

1. **New "User-Agent Configuration" section** - Added detailed explanation after the configuration options table in both English and Chinese documentation
2. **Usage scenarios** - Explained when and why to use this option
3. **Configuration examples** - Provided complete SQL examples for both lookup and sink tables
4. **Important notes** - Documented the conflict detection behavior and format recommendations

The documentation now includes:
- Clear explanation of default behavior (`flink-connector-http`)
- Real-world configuration examples showing the option in context
- Warning about not setting User-Agent in both `http.user.agent` and `http.headers`
- HTTP User-Agent format best practices

Please let me know if you'd like any additional documentation or clarification!

featzhang:
Thanks for the review. Pushed a rebuilt commit that squashes the history into a single clean commit on top of current `main` and addresses the outstanding feedback:

- Default value is `flink-connector-http` (per your earlier comment on the `getindata` identifier).
- Conflict with `http.sink.header.User-Agent` / `http.source.lookup.header.User-Agent` is now rejected at table creation time with a clear `IllegalArgumentException`, instead of silently letting one value win.
- Rebased onto `main`, so both the new `USER_AGENT` option and the `FLINK-HTTP-3` map-type header options (`SINK_HEADERS`, `SOURCE_LOOKUP_HEADERS`) are registered in the factories side by side — no regression on FLINK-HTTP-3.
- Fixed the Java 17 CI failure (missing `assertThatThrownBy` static import in the two new tests).
- Added documentation: a row in the lookup and sink options tables plus a short `User-Agent` subsection with an example DDL in both `content` and `content.zh`.

Local `mvn spotless:check` and `mvn test -pl flink-connector-http` are green (396 tests, 0 failures, 0 errors).

featzhang:
Correction on my previous comment: the force-pushed commit is now properly based on the current `apache/flink-connector-http` `main` (HEAD `31fc409`, on top of `813fd5d` `[FLINK-39364] Add request timeout`). The earlier push was accidentally rebased onto my own fork's `main`, which contained some `FLINK-HTTP-3` work that is not yet merged upstream — that's what GitHub reported as `CONFLICTING`. PR is now `MERGEABLE` and CI has been triggered.

featzhang:
Hi @davidradl, friendly ping — all three of your review threads on this PR are now marked as resolved and the latest commit (`31fc409`) addresses each of them:

| Thread | Fix |
|--------|-----|
| `JavaNetSinkHttpClient.java` L76 — *"why a dedicated config, prefix headers already work"* | Kept the dedicated `ConfigOption` so there is a named, documented entry point and a sensible default (requests no longer go out with no `User-Agent` at all). |
| `HttpDynamicSinkConnectorOptions.java` L51 — *"default to `flink-connector-http`"* | Default changed to `flink-connector-http` on both sink and lookup sides. |
| `JavaNetSinkHttpClientTest.java` L143 — *"config error on User-Agent clash"* | Setting User-Agent via both `http.user.agent` and `http.(sink|source.lookup).header.User-Agent` now throws `IllegalArgumentException` at factory creation with a clear message; covered by new unit tests on both sink and lookup clients. |

Plus: documentation rows + a `User-Agent` subsection added in `docs/content` and `docs/content.zh`, Java 17 compile fix, rebased on current `main` (`813fd5d`).

CI (GitHub Actions Java 11 + 17) is green. PR is `MERGEABLE`. Would you mind taking another look when you have a moment? Happy to iterate further if anything still looks off.

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -448,6 +450,32 @@ CREATE TABLE http-lookup (
...
+  'format' = 'json',
+  'http.user.agent' = 'MyApp/1.0'
+)
+```
+
+The same option is also available for the HTTP sink, see the [Sink Connector Options](#sink-connector-options) table.
+
+Setting `User-Agent` through both `http.user.agent` and `http.source.lookup.header.User-Agent`
+(or `http.sink.header.User-Agent` for the sink) is ambiguous and is rejected at table creation time
+with an `IllegalArgumentException`. Configure it through exactly one of the two options.
```
davidradl, line 477: nit: remove the word exactly

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -208,6 +209,7 @@ Note the options with the prefix _http_ are the HTTP connector specific options,
...
+| http.user.agent                                                        | optional | The value of the `User-Agent` HTTP header sent with every lookup request. Defaults to `flink-connector-http`. Setting this option together with `http.source.lookup.header.User-Agent` is not allowed and will fail with a configuration error.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
davidradl, line 212: I think we should put the new option near the top around url, so we start with all the common configuration options then specialize to the sink or lookup appropriately when we list them

flink-connector-http/src/main/java/org/apache/flink/connector/http/sink/httpclient/JavaNetSinkHttpClient.java
```diff
@@ -73,6 +74,23 @@ public JavaNetSinkHttpClient(
...
                         headerPreprocessor);
 
+        // Apply the User-Agent header. If the user also set it via
+        // 'http.sink.header.User-Agent', fail fast so that we do not silently drop
+        // one of the two values.
+        String userAgent =
+                properties.getProperty(
+                        HttpDynamicSinkConnectorOptions.USER_AGENT.key(),
+                        HttpDynamicSinkConnectorOptions.USER_AGENT.defaultValue());
+        if (userAgent != null && !userAgent.isEmpty()) {
```
davidradl, line 84: userAgent will always have a value here - as it will be the user specified one or the default. So there is no way to specify http.sink.header.User-Agent and have it work after this change

PR #35, opened 2026-03-15
[FLINK-39367][connector-http] Add configurable error logging with detailed context

featzhang:
## Purpose
This PR improves HTTP request error logging in the HTTP connector by adding configurable error logging with detailed context. Previously, HTTP errors were logged with minimal information, making it difficult to diagnose issues in production environments. This enhancement provides more comprehensive error details including HTTP status codes, response bodies (configurable), and configurable log levels.

## Change log
- Added `HttpErrorLogLevel` enum for configurable error log levels (ERROR/WARN/INFO)
- Added `HttpErrorLogger` class for enhanced error logging with detailed context
- Added configuration options:
  - `http.error.log.level`: Controls error log level (default: ERROR)
  - `http.error.log.include.body`: Whether to include response body in logs (default: false)
  - `http.error.log.body.max.size`: Maximum body size to log in bytes (default: 1024)
- Integrated `HttpErrorLogger` in `JavaNetHttpPollingClient` and `JavaNetSinkHttpClient`

## Verifying
- All existing tests pass
- Added comprehensive unit tests for `HttpErrorLogger` (12 tests, all passing)
- Code style checks pass (Checkstyle and Spotless)
- Manual testing confirmed:
  - Error logging works with different log levels
  - Response body inclusion is configurable
  - Body size limit is enforced
  - Logs include all relevant context information

## Affected Components
- `flink-connector-http-core`: Core HTTP client implementation
- Configuration system: New configuration parameters for error logging

## Documentation
Configuration documentation should be updated to include the new error logging parameters:
- `http.error.log.level`
- `http.error.log.include.body`
- `http.error.log.body.max.size`

## Backward Compatibility
This change is fully backward compatible. All new configuration parameters have sensible defaults, and existing code will continue to work without modifications.


flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClient.java
```diff
@@ -161,6 +164,9 @@ private HttpRowDataWrapper queryAndProcess(RowData lookupData) throws Exception
         } catch (HttpStatusCodeValidationFailedException e) {
             // log if we fail for status code reasons.
             httpLogger.logResponse((HttpResponse<String>) e.getResponse());
+            // Enhanced error logging
+            httpErrorLogger.logLookupError(
```
davidradl: why log this as an error here - when the config might have decided to continue on error. We already log the response but not as an error

flink-connector-http/src/main/java/org/apache/flink/connector/http/config/HttpConnectorConfigConstants.java
```diff
@@ -164,4 +164,15 @@ public final class HttpConnectorConfigConstants {
             SOURCE_RETRY_EXP_DELAY_PREFIX + "max-backoff";
     public static final String SOURCE_RETRY_EXP_DELAY_MULTIPLIER =
             SOURCE_RETRY_EXP_DELAY_PREFIX + "backoff-multiplier";
+
+    // -------------- Error logging configuration --------------
+
+    public static final String ERROR_LOG_LEVEL = FLINK_CONNECTOR_HTTP + "error.log.level";
```
davidradl: We already have 'http.logging.level' that dictate how much should be logged. This includes the body in some cases. 

could you raise a Jira on this one so we can see the motivation behind this change. My thought is that is should extend or enhance the existing mechanism rather than introduce a new logging mechanism

featzhang: Thanks for the review feedback. I've raised a Jira to document the motivation behind this change.

  To clarify why this introduces a new mechanism rather than extending http.logging.level:

  The existing http.logging.level controls content verbosity (MIN/REQ_RESP/MAX) and is always emitted at DEBUG level. It is designed for development-time
  debugging and logs every request and response regardless of outcome. This creates two gaps for production use:

  1. In production, DEBUG logging is typically disabled. Errors become invisible unless DEBUG is enabled for the entire connector, which floods logs with
  all successful requests as well.
  2. There is no way to emit error information at ERROR or WARN level so that log-based alerting systems (Splunk, Datadog, etc.) can detect failures.

  The new http.error.log.* configuration addresses a different and orthogonal concern: error severity. A user may want minimal content (no body) but still
  want errors surfaced at ERROR level for alerting. Or they may want MAX content detail but only at DEBUG. These two dimensions — content verbosity and
  error severity — are independent and should be configured separately.

  One important behavioral detail: HttpErrorLogger respects continue-on-error. When continue-on-error=true, errors are tolerated and only a DEBUG message
  is emitted, since logging at ERROR/WARN would be misleading. The configured error log level only applies when the exception is actually thrown.

  I've also split the concerns in the Jira description to make the motivation clearer. Happy to discuss further if you have concerns about the approach.

featzhang: Thanks for the feedback. I've raised a Jira to document the motivation: https://issues.apache.org/jira/browse/FLINK-39367.

  To clarify the reasoning behind introducing a separate mechanism rather than extending http.logging.level:

  The existing http.logging.level controls content verbosity — how much detail (headers, body) is logged for every HTTP request and response. It is always emitted at DEBUG level and is intended for development-time troubleshooting.

  The new http.error.log.* configuration addresses a different concern: error severity — at what SLF4J level (ERROR/WARN/INFO) failures are surfaced. These are orthogonal dimensions:

  - A user may want minimal content (no body) but still want errors logged at ERROR level so that alerting systems can detect failures. With only http.logging.level, this is impossible because all output is DEBUG.
  - Conversely, a user may want MAX content detail but only at DEBUG to avoid noise in monitoring dashboards.

  Extending http.logging.level to also control log severity would conflate two unrelated concerns into a single option, making the configuration harder to reason about. Keeping them separate matches the pattern used in other Flink connectors where error handling and general logging are configured independently.

  One additional behavioral detail worth noting: HttpErrorLogger respects continue-on-error. When continue-on-error=true, errors are tolerated and only a DEBUG message is emitted — logging at ERROR/WARN in that case would be misleading since the connector is functioning as configured. The configured error log level only applies when the exception will actually be thrown.

davidradl: I think we should fit in with the [existing configuration](https://github.com/apache/flink-connector-http/blob/main/docs/content/docs/connectors/table/http.md#logging-the-http-content) based logging . We added this so that potentially sensitive information in the http messages could be obfuscated.  We agreed that it would be possible to log out the clear content for development debugging. I suggest we use the existing controls and scope and extend them to to cover sink

PR #36, opened 2026-03-16
[FLINK-39235] Simplify body processing for GenericJsonAndUrlQueryCreator

davidradl:


flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -281,6 +297,85 @@ static String convertToQueryParameters(final ObjectNode jsonObject, String enc)
...
+        // If source array is empty or target array is empty, replace with source
+        if (sourceArray.size() == 0 || targetArray.size() == 0) {
+            target.set(fieldName, sourceArray);
+            return;
+        }
+
+        // Check if source array contains objects (use first element as template)
+        JsonNode sourceTemplate = sourceArray.get(0);
+        if (!sourceTemplate.isObject()) {
+            // For primitive arrays, replace entire array with source
```
davidradl: this is not correct - will fix

davidradl:
@ferenc-csaky fyi I spotted an error - will correct - then put this back to ready for review.

seb-pereira:
**Expand Java prerequisites to Java 11-17?**

Not specific to changes from this PR but when I built it locally I checked the [README](https://github.com/apache/flink-connector-http?tab=readme-ov-file#building-the-apache-flink-http-connector-from-source) that currently lists Java 11 as a prerequisite, which may give the impression that newer Java versions are not supported.

This project uses Flink 1.20, which officially supports Java 11 and 17 according to the [Flink documentation](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/deployment/java_compatibility/).

I tested with Java 17 and confirm the build succeeded without any issues. 

@davidradl could you consider updating the README to reflect Java 11-17 support?

---

For informational purposes: I also tried to test building with Java 21 -  / [documented for 2.2.0](https://nightlies.apache.org/flink/flink-docs-release-2.2/docs/deployment/java_compatibility/) and official [Flink 2.2 Docker images use Java 21](https://github.com/apache/flink-docker/blob/fbe91e0c410a77c7c894786fdf9d118b804eb2e7/2.2/scala_2.12-java21-ubuntu/Dockerfile)), and found only one issue when building the project:
   ```
  java.lang.reflect.InvocationTargetException
  Caused by: java.lang.NoSuchMethodError: 'com.sun.tools.javac.tree.JCTree com.sun.tools.javac.tree.JCTree$JCImport.getQualifiedIdentifier()'
  ```
=> The inherited Spotless Maven Plugin (2.27.1) with Google Java Format (1.7) does not support Java 21.

For those interested in building with Java 21, the following changes enable compatibility:

```xml
<properties>
    <spotless.version>2.43.0</spotless.version>
    <spotless.google-java-format.version>1.17.0</spotless.google-java-format.version>
    <spotless.skip>false</spotless.skip>
</properties>

<plugin>
    <groupId>com.diffplug.spotless</groupId>
    <artifactId>spotless-maven-plugin</artifactId>
    <version>${spotless.version}</version>
    <configuration>
        <java>
            <googleJavaFormat>
                <version>${spotless.google-java-format.version}</version>
            </googleJavaFormat>
        </java>
    </configuration>
</plugin>
```

davidradl:
@seb-pereira could you raise a separate issue for the java version.

davidradl:
@seb-pereira @ferenc-csaky 
I am looking to be able to replace sections of the payload with literals. I want to be able to specify a json path and a value that will be put at that path. The path can be an object, array or literal. As such the existing design does not handle this because an object 

"/obj1/str1" = "aaa" 

could mean we want to replace str1 with "aaa" or to replace "obj1" with str1="aaa" 

In order to be able to handle these cases, we looked for standard practise and found [JSON patch](https://jsonpatch.com/). Which allows 
"
The original document
{
  "baz": "qux",
  "foo": "bar"
}
The patch
[
  { "op": "replace", "path": "/baz", "value": "boo" },
  { "op": "add", "path": "/hello", "value": ["world"] },
  { "op": "remove", "path": "/foo" }
]
"

This is useful but if the field is in an object that does not exist in the event stream, then we want to force the literal in, 
[Stratgic merge](https://www.linkedin.com/pulse/patches-kustomize-json-6902-vs-strategic-merge-patch-adamson-hwngc/) allows this. But is a merge.

We are wanting a Strategic replace, so it forces the literal value into the path, whether or not there is a parent object.

Inspired by these approaches; we are looking at a payload of :
```
[
  { "path": "/baz", "value": "boo" },
   ...
]

```

davidradl:
After further investigation we are now thinking there is a better way that would make the population of the body much more intuitive for this query creator. The idea is to remove `http.request.additional-body-json` and `http.request.body-fields` and add `http.request.body-template`. This will contain the body that we want to send 
‘
{
       “Id” : <customerId>,
       “Orders” : {
   				“orderId”: <topOrderId>,
				“version:  “1.0”,
                                “Address” : {
 					“streetName”: <topStreetName>,
 					“City”:  <topCity>,
 	                                “Country” : “UK”  
                                 }
          } 
}
‘ 
where the angled bracket names are top level columns. The join keys must only reference top level lookup table columns, but may need to be put into a nested part of the body.

Prototyping this; as it appears to be a way to cover all cases in a reasonably simple and intuitive manner, without the need for path mapping.

davidradl:
@ferenc-csaky are you ok to have a look at this . The HTTP body processing has been massively simplified and is driven by one new configuration option. I have removed the others. Many thanks

ferenc-csaky:
@seb-pereira @davidradl  FYI If we wanna release the connector for Flink 1.20 (IMO we should), JDK8 cannot be dropped yet.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -178,6 +162,42 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
...
+     *
+     * @param template the template string with placeholders
+     * @param jsonObject the JSON object containing field values
+     * @return the template with placeholders replaced by actual values
+     */
+    private String substituteTemplate(String template, ObjectNode jsonObject) {
+        java.util.regex.Pattern pattern = java.util.regex.Pattern.compile("<([^>]+)>");
+        java.util.regex.Matcher matcher = pattern.matcher(template);
+
+        StringBuffer result = new StringBuffer();
```
ferenc-csaky: Why `StringBuffer`? `StringBuilder` should always be the default choice, cause it's thread-safe, unless specific concurrency requirements.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -178,6 +162,42 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
...
+     * Substitutes placeholders in the template with values from the JSON object. Placeholders are
+     * in the format {@code <fieldName>} where fieldName is a top-level field in the JSON object.
+     *
+     * @param template the template string with placeholders
+     * @param jsonObject the JSON object containing field values
+     * @return the template with placeholders replaced by actual values
+     */
+    private String substituteTemplate(String template, ObjectNode jsonObject) {
+        java.util.regex.Pattern pattern = java.util.regex.Pattern.compile("<([^>]+)>");
+        java.util.regex.Matcher matcher = pattern.matcher(template);
```
ferenc-csaky: nit: Let's import these regex classes.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -178,6 +162,42 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
...
 
+    /**
+     * Substitutes placeholders in the template with values from the JSON object. Placeholders are
+     * in the format {@code <fieldName>} where fieldName is a top-level field in the JSON object.
+     *
+     * @param template the template string with placeholders
+     * @param jsonObject the JSON object containing field values
+     * @return the template with placeholders replaced by actual values
+     */
+    private String substituteTemplate(String template, ObjectNode jsonObject) {
```
ferenc-csaky, line 175: Any specific reason for the `<variableName>` placeholder syntax? AFAIK this is not something that any widely used JSON templating engine uses, so if no reason let's not create yet another syntax. I'd rather go with the Mustache/Handlebars syntax: `{{variableName}}`, that's the most popular IMO.

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -163,50 +163,49 @@ Or for REST POST method they will be converted to Json and used as request body.
...
+| http.request.body-template                                             | optional | Used for the `http-generic-json-url` query creator. A JSON template string for constructing the request body for PUT and POST operations. Use `<fieldName>` placeholders to reference top-level columns from the lookup table. Supports creating complex nested JSON structures with both placeholders and literal values. Example: `'{"user": {"id": <userId>, "name": <userName>}, "source": "flink"}'`. See the [Body Template](#body-template) section for details and examples.                                                                                                                          |
```
ferenc-csaky: Since we explicitly refer the detailed section here, I think we may remove the example from here.

ferenc-csaky (changes requested):

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
davidradl: Fair point, I did consider it but chose <> because json has curly brackets . I will amend.

davidradl:
> @seb-pereira @davidradl FYI If we wanna release the connector for Flink 1.20 (IMO we should), JDK8 cannot be dropped yet.

We are now only interested in version 2 and the current connector is built at 1.20 and has only ever supported java 11. I will investigate how easy it is to downgrade the java. If it is looking not straight forward - we may want to go straight to v2.

davidradl:
@ferenc-csaky changes addressed - back with you

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -171,8 +173,8 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
      * @return the template with placeholders replaced by actual values
      */
     private String substituteTemplate(String template, ObjectNode jsonObject) {
-        java.util.regex.Pattern pattern = java.util.regex.Pattern.compile("\\{\\{([^}]+)\\}\\}");
-        java.util.regex.Matcher matcher = pattern.matcher(template);
+        Pattern pattern = java.util.regex.Pattern.compile("\\{\\{([^}]+)\\}\\}");
```
ferenc-csaky: `Pattern.compile` is still uses the full pkg.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -164,26 +164,29 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
...
-                throw new IllegalArgumentException(
-                        String.format(
-                                "Template placeholder <%s> references a field that does not exist in the lookup row",
-                                fieldName));
+                IllegalArgumentException illegalArgumentException =
+                        new IllegalArgumentException(
+                                String.format(
+                                        "Template placeholder {{%s}} references a field that does not exist in the lookup row",
+                                        fieldName));
+                log.error(illegalArgumentException.getMessage(), illegalArgumentException);
```
ferenc-csaky: Any specific reason to explicitly log here too?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
davidradl: sorry missed this

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
davidradl: I was not seeing this exception  in the log - I want to check this separately to ensure the message does make it to the log. Removing for now to get this through.

davidradl:
@ferenc-csaky back to you for review - raised [FLINK-39332](https://issues.apache.org/jira/projects/FLINK/issues/FLINK-39332) to ensure that the Exception error message is in the log

ferenc-csaky (approved):
LGTM, thanks!

PR #37, opened 2026-03-27
[hotfix] Fix SELECT statement in http.md

sofie204:
The `SELECT` example in the HTTP lookup source docs uses `c.isActive` and `c.balance`, but the table schema defines these fields nested inside a details ROW. This causes NULL results when users follow the example. The fix is to use `c.details.isActive` and `c.details.nestedDetails.balance`.

### DIFF
--- a/docs/content/docs/connectors/table/http.md
+++ b/docs/content/docs/connectors/table/http.md
@@ -140,7 +140,7 @@ CREATE TABLE Orders (
 Then we can enrich the _Orders_ table with the _Customers_ HTTP table with the following SQL:
 
 ```roomsql
-SELECT o.id, o.id2, c.msg, c.uuid, c.isActive, c.balance FROM Orders AS o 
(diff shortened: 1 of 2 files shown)


docs/content/docs/connectors/table/http.md
```diff
@@ -140,7 +140,7 @@ CREATE TABLE Orders (
 Then we can enrich the _Orders_ table with the _Customers_ HTTP table with the following SQL:
 
 ```roomsql
-SELECT o.id, o.id2, c.msg, c.uuid, c.isActive, c.balance FROM Orders AS o
```
davidradl, line 143: could you make the changes in the chinese folder as well please. I forgot to mention this earlier.

davidradl:
nit: the title should be statement not statment

sofie204:
> nit: the title should be statement not statment

Thank you for pointing that out @davidradl. Both changes have been made and the spelling has been modified.

PR #38, opened 2026-03-31
[FLINK-39333] add query params key

davidradl:


flink-connector-http/src/test/java/org/apache/flink/connector/http/app/AbstractJsonTransform.java
```diff
@@ -18,27 +18,30 @@
...
+ * <p>Subclasses must provide:
+ *
+ * <ul>
+ *   <li>Transformer name via {@link #getName()}
+ *   <li>JSON template via {@link #getJsonTemplate()}
+ *   <li>UUID generation strategy via {@link #getUuidValue()}
+ * </ul>
  */
-public class JsonTransform extends ResponseTransformer {
+public abstract class AbstractJsonTransform implements ResponseTransformerV2 {
```
davidradl, line 40: ResponseTransformer was deprecated so I updated this to use the recommended interface. I also refactored to avoid duplicated code and gave the json transforms unque names.

davidradl:
@ferenc-csaky please could you process this one?

davidradl:
@rmetzger As discussed please could you have a look at this one. This is blocking us releasing the connector

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
```diff
@@ -136,7 +162,138 @@ public Set<ConfigOption<?>> requiredOptions() {
...
+                throw new IllegalArgumentException(
+                        "Query parameter configuration ("
+                                + REQUEST_QUERY_PARAM_FIELDS.key()
+                                + " or "
+                                + REQUEST_QUERY_PARAM_FIELDS_WITH_KEY.key()
+                                + ") can only be used with GET method. "
+                                + "For POST/PUT requests, use "
+                                + REQUEST_BODY_TEMPLATE.key()
+                                + " instead.");
+            }
```
rmetzger, line 295: My AI tells me this:

The new `validateConfigurationForHttpMethod()` throws `IllegalArgumentException` when `http.request.query-param-fields` is set with POST or PUT. Previously, this was valid — query param fields were used to populate `bodyBasedUrlQueryParams` for the request body. Evidence:
- The old `getConfiguration()` test helper set `REQUEST_QUERY_PARAM_FIELDS` for ALL operations including POST/PUT
- The old `validateCreatedQueryForPutAndPost` asserted `hasBodyBasedUrlQueryParameters().isTrue()`

**Any existing user** with a config like `'lookup-method' = 'POST', 'http.request.query-param-fields' = 'id;name'` will get a runtime error after upgrading with no migration path other than rewriting to use `body-template`.

**Recommendation:** Either remove the validation to preserve backward compatibility, or document this as an intentional breaking change with migration guidance.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
```diff
@@ -132,9 +134,15 @@ public LookupQueryInfo createLookupQuery(final RowData lookupDataRow) {
...
         ObjectNode jsonObjectForQueryParams = ObjectMapperAdapter.instance().createObjectNode();
-        for (String requestColumnName : this.requestQueryParamsFields) {
-            jsonObjectForQueryParams.set(requestColumnName, jsonObject.get(requestColumnName));
+        for (Map.Entry<String, String> entry : this.requestQueryParamsMap.entrySet()) {
+            String columnName = entry.getKey();
+            String queryParamKey = entry.getValue();
+            JsonNode value = jsonObject.get(columnName);
+            if (value != null) {
+                jsonObjectForQueryParams.set(queryParamKey, value);
+            }
```
rmetzger, line 148: My AI tells me:

```java
JsonNode value = jsonObject.get(columnName);
if (value != null) {
    jsonObjectForQueryParams.set(queryParamKey, value);
}
```

If a user configures `'http.request.query-param-fields-with-key' = 'misspelledCol:customer'`, the query parameter is **silently omitted** from the HTTP request. No error, no warning. Compare this to `substituteTemplate()` at line 191 which correctly throws `IllegalArgumentException` for missing fields — the two code paths have inconsistent error handling for the same class of problem.

**Recommendation:** Throw an `IllegalArgumentException` with the misconfigured column name and available columns, consistent with `substituteTemplate()`.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
davidradl: As this has not be released I do not think that we need to be concerned with backward compatibility. I will make a change to the docs to highlight this for migration.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
davidradl: sounds like a good change (with a unit test) - my AI did not spot this when I asked it to review !

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreatorFactory.java
ferenc-csaky: Regarding David's point, I think it's fine to change this and leave the logic as is.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/querycreators/GenericJsonAndUrlQueryCreator.java
ferenc-csaky: +1 for this addition. If this gets covered, the rest LGTM.

davidradl:
@rmetzger I have made the changes. Are you happy for me to merge. @ferenc-csaky has LGTM.

ferenc-csaky (approved):

PR #39, opened 2026-04-24
[FLINK-39506] Build with Flink version 2

davidradl:


davidradl:
@ferenc-csaky I have moved the connector up to build on Version 2. Please could you have a look.

.github/workflows/weekly.yml
```diff
@@ -29,13 +29,26 @@ jobs:
...
+          branch: release-2.2
         }]
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
       flink_version: ${{ matrix.flink_branches.flink }}
       connector_branch: ${{ matrix.flink_branches.branch }}
-      jdk_version: ${{ matrix.flink_branches.jdk || '11, 17' }}
+      jdk_version: ${{ matrix.flink_branches.jdk || '11, 17, 21' }}
       run_dependency_convergence: false
+  python_test:
```
ferenc-csaky: The connector does not have a python pkg, so IMO we should omit `python_test`.

flink-connector-http/pom.xml
```diff
@@ -211,6 +205,14 @@ under the License.
             <artifactId>mockito-inline</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-runtime</artifactId>
+            <version>${flink.version}</version>
+            <scope>provided</scope>
+        </dependency>
```
ferenc-csaky, line 214: I'd define this in the root POM's `dependencyManagement` section, and only reference with scope but without version.

flink-connector-http/src/test/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClientWithWireTest.java
```diff
@@ -156,7 +154,7 @@ private void shouldUpdateHttpRequestIfRequired(HttpRequest httpRequest)
         configuration.setString(
                 SOURCE_LOOKUP_OIDC_AUTH_TOKEN_ENDPOINT_URL.key(),
                 "http://localhost:" + SERVER_PORT + "/auth");
-        configuration.setString(SOURCE_LOOKUP_OIDC_AUTH_TOKEN_REQUEST, BEARER_REQUEST);
+        configuration.setString(SOURCE_LOOKUP_OIDC_AUTH_TOKEN_REQUEST.key(), BEARER_REQUEST);
```
ferenc-csaky: nit: 
```java
configuration.set(SOURCE_LOOKUP_OIDC_AUTH_TOKEN_REQUEST, BEARER_REQUEST);
```

flink-connector-http/src/test/java/org/apache/flink/connector/http/table/lookup/JavaNetHttpPollingClientConnectionTest.java
```diff
@@ -292,9 +292,9 @@ void shouldHandleCodeBasedOnConfiguration(
 
         // GIVEN
         this.stubMapping = setUpServerStub(201);
-        configuration.setString(SOURCE_LOOKUP_HTTP_SUCCESS_CODES, successCodesExpression);
+        configuration.setString(SOURCE_LOOKUP_HTTP_SUCCESS_CODES.key(), successCodesExpression);
```
ferenc-csaky: All of `configuration.setString` should be migrated to `configuration.set` instead. Then you can keep passing the `ConfigOption` object as it was before.

flink-connector-http/src/test/java/org/apache/flink/connector/http/table/lookup/BodyBasedRequestFactoryTest.java
```diff
@@ -49,9 +49,9 @@ void testconstructUri(TestSpec testSpec) throws Exception {
         Configuration configurationHttp2 = new Configuration();
 
         configurationHttp2.setString(
-                LOOKUP_HTTP_VERSION, String.valueOf(HttpClient.Version.HTTP_2));
+                LOOKUP_HTTP_VERSION.key(), String.valueOf(HttpClient.Version.HTTP_2));
```
ferenc-csaky: Change `setString` to `set` instead of passing key.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/AsyncHttpTableLookupFunction.java
```diff
@@ -126,6 +126,7 @@ public HttpLookupConfig getOptions() {
     public void close() throws Exception {
         this.publishingThreadPool.shutdownNow();
         this.pullingThreadPool.shutdownNow();
+        decorate.close();
```
ferenc-csaky, line 129: This was just missing before, correct?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/RequestFactoryBase.java
```diff
@@ -51,7 +51,7 @@ public abstract class RequestFactoryBase implements HttpRequestFactory {
     /** HTTP headers that should be used for {@link HttpRequest} created by factory. */
     private final String[] headersAndValues;
 
-    private final HttpLookupConfig options;
+    private final transient HttpLookupConfig options;
```
ferenc-csaky: Why making it transient here. I'm not saying it's wrong, I just don't understand from the limited context of this PR, cause the most significant change in making it sure this object is properly serializable. :)

flink-connector-http/src/main/java/org/apache/flink/connector/http/utils/JavaNetHttpClientFactory.java
```diff
@@ -67,30 +67,34 @@ public static HttpClient createClient(HttpLookupConfig options) {
...
-                    readableConfig.getOptional(
-                            HttpLookupConnectorOptions.SOURCE_LOOKUP_PROXY_PASSWORD);
-
-            ProxyConfig proxyConfig =
-                    new ProxyConfig(proxyHost.get(), proxyPort.get(), proxyUsername, proxyPassword);
-            clientBuilder.proxy(
-                    ProxySelector.of(
-                            new InetSocketAddress(proxyConfig.getHost(), proxyConfig.getPort())));
-            proxyConfig.getAuthenticator().ifPresent(clientBuilder::authenticator);
+        if (readableConfig != null) {
```
ferenc-csaky: I'm not sure I see why this guard is needed now and was not necessary before. Same as the null-check and exception throw at the `JavaNetHttpPollingClient` ctor.

ferenc-csaky (changes requested):
Thanks, added raised some comments and questions.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/AsyncHttpTableLookupFunction.java
davidradl: yes - this was cleaning up . Prior to this change the aync tests were leaking into each other this was part of cleaning that up

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/RequestFactoryBase.java
davidradl: I added this as part of debugging - and did not remove - removing

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpTableLookupFunction.java
```diff
@@ -111,13 +117,59 @@ public Collection<RowData> lookup(RowData keyRow) {
             physicalArity = physicalRow.getArity();
             producedRow =
                     new GenericRowData(physicalRow.getRowKind(), physicalArity + metadataArity);
-            // We need to copy in the physical row into the producedRow
+            // Build a map of lookup table field names to their typed values from keyRow
+            // Only process top-level single-value join keys
+            Map<String, Object> joinKeyValues = new HashMap<>();
```
davidradl, line 122: note this was a runtime regression that started occurring at Flink 2 where non null join keys were being returned as null from the connector - in this case we put in the join key value with joined on.

davidradl:
@ferenc-csaky I have made the requested changes . I also fixed a runtime regression we hit - there is comment I opened to explain it to you.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceFactory.java
```diff
@@ -220,12 +221,16 @@ private HttpLookupConfig getHttpLookupOptions(Context context, ReadableConfig re
...
+        Configuration configuration = new Configuration();
+        readableConfig.toMap().forEach(configuration::setString);
+
         return HttpLookupConfig.builder()
                 .lookupMethod(readableConfig.get(LOOKUP_METHOD))
                 .url(readableConfig.get(URL))
                 .useAsync(readableConfig.get(ASYNC_POLLING))
                 .properties(httpConnectorProperties)
-                .readableConfig(readableConfig)
+                .readableConfig(configuration)
```
ferenc-csaky: We should be able to simplify this to:
```java
.readableConfig(Configuration.fromMap(readableConfig.toMap()))
```
It does the same thing inside that now happens in L225-226

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpTableLookupFunction.java
ferenc-csaky: so this was not happening until the connector was built with Flink 1.20? even when runtime was 2.0? i'd like to understand what's going on here.

davidradl: @ferenc-csaky I have done some more testing , prior to Flink 1.20 the checking was not as rigourous. I have tested the main version again and see the issue at 1.20; I back ported this change to my local 1.20 and the fix made it work again. This was a false alarm. Are you ok for this fix to remain in this PR?

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceFactory.java
davidradl: I have implemented this change , but checked whether we already were a configuration in which case we do not need to do this to and from map.

davidradl:
Hi @ferenc-csaky back with you for review

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/HttpTableLookupFunction.java
ferenc-csaky: Yep, I think that's okay. Thanks for looking into it.

ferenc-csaky (approved):
LGTM

PR #40, opened 2026-05-14
[FLINK-39675][connector-http] remove query param config

davidradl:


ferenc-csaky (approved):
Change LGTM. As we discussed offline with @davidradl, we should align the templating syntax, as now we `{myVar}` and `{{myVar}}` for 2 config options. That is planned to be done in a separate upcoming change.

PR #43, opened 2026-05-20
[hotfix] Amend title in docs

davidradl:
amended the title in the docs to be consistent with the Kafka connector

### DIFF
--- a/docs/content.zh/docs/connectors/table/http.md
+++ b/docs/content.zh/docs/connectors/table/http.md
@@ -25,7 +25,7 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-# HTTP Connector
+# Apache HTTP SQL Connector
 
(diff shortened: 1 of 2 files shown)


ferenc-csaky (changes requested):
I don't think this change is meaningful or should happen at all, so I am -1 on this. There is nothing Apache about HTTP. And not all connectors have that prefix (e.g. `HBase SQL Connector`, `Upsert Kafka SQL Connector` etc).

davidradl:
> I don't think this change is meaningful or should happen at all, so I am -1 on this. There is nothing Apache about HTTP. And not all connectors have that prefix (e.g. `HBase SQL Connector`, `Upsert Kafka SQL Connector` etc).

Ok I had just seen the Kafka one . Happy not to change this.

PR #49, opened 2026-06-16
[hotfix] Notice change

davidradl:
To address @rmetzger 's [comment](https://lists.apache.org/thread/y0y1g6kh0fot7mj6ft40cz16wx0hl89c) on the feedback on release 1.0.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -644,10 +644,12 @@ under the License.
                             <ignoredDependencies>
                                 <!--These dependencies are required for the unit tests to succeed  -->
                                 <ignoredDependency>org.mockito:mockito-inline:jar:4.6.1</ignoredDependency>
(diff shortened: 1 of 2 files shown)


davidradl:
@rmetzger can you confirm this address the issue you raised please

pom.xml
nictownsend, line 648: What was the purpose of moving this around?

I also note that `flink-clients` has no scope, and `flink-streaming-java` has scope `provided` -should both of them be `test` scope?

nictownsend: Alternatively, if this was an artifact from other work, I guess we should revert and make a new commit for pom change (if it's needed)

nictownsend (changes requested):

pom.xml
davidradl: this as inadvertently added - thanks for spotting it

PR #31, opened 2026-03-11
[FLINK-39364][connector-http] Add request timeout config option

featzhang:
## What is the purpose of the change

The HTTP request timeout for the sink (`http.sink.request.timeout`) and lookup source (`http.source.lookup.request.timeout`) were already supported at runtime, but were not exposed as typed `ConfigOption` definitions in their respective connector options classes.

This meant:
1. The options were not listed in `optionalOptions()` of the `DynamicTableFactory` implementations — Flink could not validate them.
2. Users received a confusing "unsupported option" warning when configuring them via SQL DDL.
3. There was no type safety, no default-value documentation, and no IDE auto-completion support.
4. A `TODO` comment existed in `HttpLookupTableSourceFactory` noting this gap.

This PR resolves the above by adding proper `ConfigOption<Duration>` definitions and registering them in their factories.

## Brief change log

- Add `HttpLookupConnectorOptions.SOURCE_LOOKUP_REQUEST_TIMEOUT`
  - Key: `http.source.lookup.request.timeout`
  - Type: `Duration`
  - Default: `30s`
- Add `HttpDynamicSinkConnectorOptions.SINK_REQUEST_TIMEOUT`
  - Key: `http.sink.request.timeout`
  - Type: `Duration`
  - Default: `30s`
- Register both options in `optionalOptions()` of their respective `DynamicTableFactory`
- Remove the `TODO` comment from `HttpLookupTableSourceFactory`
- Add unit tests in `HttpLookupTableSourceFactoryTest` and `HttpDynamicTableSinkFactoryTest`

## Verifying this change

- New unit tests in `HttpLookupTableSourceFactoryTest`:
  - `shouldAcceptWithRequestTimeout`: verifies that `http.source.lookup.request.timeout=60s` is accepted without a `ValidationException`
  - `shouldUseDefaultRequestTimeoutWhenNotConfigured`: verifies that omitting the option still creates the source successfully
- New unit test in `HttpDynamicTableSinkFactoryTest`:
  - `requestTimeoutOptionTest`: verifies that `http.sink.request.timeout=60s` is accepted as a valid option

All 11 affected tests pass. Full build with `mvn clean install -DskipTests -Pfast` succeeds.

## Does this pull request potentially affect one of the following areas?

- **Dependencies**: No
- **Performance**: No
- **Security**: No
- **Breaking changes**: No — this is purely additive. The runtime behaviour is unchanged; only the option declaration is formalised.

## Documentation

[...]

PR #52, opened 2026-07-04
[FLINK-40072][connector-http] Read request.timeout as the declared Duration option

bharathgunapati:

`request.timeout` is declared as a typed `ConfigOption<Duration>` for both the
lookup source (`SOURCE_LOOKUP_REQUEST_TIMEOUT`) and the sink
(`SINK_REQUEST_TIMEOUT`), defaulting to 30s, and the docs describe it as a Flink
`Duration`. However, the runtime never read the declared option — it parsed the
raw config string with `Integer.parseInt(...)`, so any `Duration`-formatted value
such as `'30s'` threw `NumberFormatException` at query time.

The failure is easy to miss because the `http.*` namespace is skipped by
`FactoryUtil#validateExcept(...)`: `CREATE TABLE` succeeds and the job only
crashes when the first lookup/sink request is built.

This completes FLINK-39364 (#31), which added the `Duration` option declarations
and docs but left the runtime read path parsing integer seconds.

## Brief change log

- Lookup: `RequestFactoryBase` reads `SOURCE_LOOKUP_REQUEST_TIMEOUT` from
  `ReadableConfig` as a `Duration` and applies it directly.
- Sink: `AbstractRequestSubmitter` parses the value with
  `TimeUtils.parseDuration(...)`; `PerRequestSubmitter` / `BatchRequestSubmitter`
  use the `Duration` directly.
- Removed the now-unused integer-seconds field and `DEFAULT_REQUEST_TIMEOUT_SECONDS`
  constant; the default comes from the option (`Duration.ofSeconds(30)`).
- Updated the DataStream docs (en + zh) to describe the value as a Flink `Duration`.

## Compatibility note

Values are now parsed with standard Flink `Duration` semantics, consistent with
the sibling `connection.timeout` option:

- Unit-suffixed values (`'30s'`, `'1min'`, `'500ms'`) now work (previously crashed).
- A bare number such as `'30'` is now interpreted as **30 milliseconds** (Flink's
  default duration unit), whereas the old broken code treated `'30'` as 30 seconds.

The only overlap is bare integers: anyone who wrote `'30'` meaning seconds should
now write `'30s'`. Any unit-suffixed value was previously non-functional (it
threw), so no working configuration silently changes meaning. This is documented
and captured in the JIRA release note.

## Verifying this change

Automated tests (Flink 2.2.0, JDK 11/17/21, CI green):
- `HttpLookupTableSourceITCaseTest#testHttpLookupJoinWithDurationRequestTimeout` —
  end-to-end lookup join with `'http.source.lookup.request.timeout' = '30s'`;
  fails on `main` with `NumberFormatException`, passes here.
- `BodyBasedRequestFactoryTest` / `BatchRequestSubmitterTest` — pin the `Duration`
  parsing semantics (units, default 30s, and the bare-number-is-milliseconds rule).
- `HttpDynamicTableSinkFactoryTest#requestTimeoutOptionTest` — a `Duration`-typed
  sink timeout is accepted.

Stock reproduction (optional, for reviewers): a single test on unmodified pre-fix
code is available at
`bharathgunapati/flink-connector-http@demo/FLINK-40072-request-timeout-repro`
(`HttpLookupTableSourceITCaseTest#testHttpLookupJoinWithDurationRequestTimeoutFailsOnPreFixRuntime`),
which asserts the `NumberFormatException` end users hit at query time.

- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: **no**
- The serializers: **no**
- The runtime per-record code paths (performance sensitive): **no**
- The S3 file system connector: **no**

## Documentation

- Does this pull request introduce a new feature? **no** (bug fix)
- If yes, how is the feature documented? docs updated to reflect the `Duration` format


flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/RequestFactoryBase.java
```diff
@@ -44,14 +44,12 @@
 @Slf4j
 public abstract class RequestFactoryBase implements HttpRequestFactory {
 
-    public static final String DEFAULT_REQUEST_TIMEOUT_SECONDS = "30";
```
bharathgunapati, line 47: Removed the hardcoded `DEFAULT_REQUEST_TIMEOUT_SECONDS = "30"` fallback.

The default now comes from the declared option itself:
`HttpLookupConnectorOptions.SOURCE_LOOKUP_REQUEST_TIMEOUT.defaultValue()`
(`Duration.ofSeconds(30)`), which is the single source of truth on
`SOURCE_LOOKUP_REQUEST_TIMEOUT`.

When the user does not set `http.source.lookup.request.timeout`, we read it via
`readableConfig.get(SOURCE_LOOKUP_REQUEST_TIMEOUT)`; if `ReadableConfig` is
unavailable (e.g. certain test/explain paths), we fall back to the option’s
`.defaultValue()` rather than duplicating `"30"` in runtime code.

This avoids two defaults drifting apart (string constant vs. `ConfigOption`) and
matches how other Flink connectors resolve typed options.

flink-connector-http/src/main/java/org/apache/flink/connector/http/sink/httpclient/AbstractRequestSubmitter.java
```diff
@@ -53,11 +55,14 @@ public AbstractRequestSubmitter(
...
                                 ThreadUtils.LOGGING_EXCEPTION_HANDLER));
 
-        this.httpRequestTimeOutSeconds =
-                Integer.parseInt(
-                        properties.getProperty(
-                                HttpConnectorConfigConstants.SINK_HTTP_TIMEOUT_SECONDS,
-                                DEFAULT_REQUEST_TIMEOUT_SECONDS));
+        // Sink runtime is wired with Properties today; reading SINK_REQUEST_TIMEOUT from
+        // ReadableConfig will be handled in a follow-up sink config-unification PR.
+        String requestTimeout =
```
bharathgunapati, line 60: Heads-up on scope: this PR fixes `request.timeout` parsing on the sink path
(`TimeUtils.parseDuration(...)` instead of `Integer.parseInt(...)`), but the
value is still sourced from the `Properties` here rather than from the typed
`SINK_REQUEST_TIMEOUT` option via `ReadableConfig`.

That asymmetry is intentional for this bug fix — the sink runtime submitter is
only handed `Properties` today (`RequestSubmitterFactory#createSubmitter`).
Reading `SINK_REQUEST_TIMEOUT` from `ReadableConfig` and threading typed config
through `HttpSinkInternal` → `RequestSubmitterFactory` → `AbstractRequestSubmitter`
will be handled in a separate sink config-unification follow-up PR.

bharathgunapati:
Hi @davidradl — would you be able to review this when you have a moment?

This is a follow-up to FLINK-39364 (#31): `request.timeout` was declared as a
`ConfigOption<Duration>` but the runtime still parsed it with
`Integer.parseInt(...)`, so values like `'30s'` crashed at query time with
`NumberFormatException`. This PR wires the runtime to the declared option on
both lookup and sink paths.

A few things that may help review:
- **JIRA:** [FLINK-40072](https://issues.apache.org/jira/browse/FLINK-40072) (includes release note for the `'30'` → `'30s'` migration)
- **Optional repro (my fork, not Apache):**
  - Branch: [demo/FLINK-40072-request-timeout-repro](https://github.com/bharathgunapati/flink-connector-http/tree/demo/FLINK-40072-request-timeout-repro)
  - Test: [`testHttpLookupJoinWithDurationRequestTimeoutFailsOnPreFixRuntime`](https://github.com/bharathgunapati/flink-connector-http/blob/demo/FLINK-40072-request-timeout-repro/flink-connector-http/src/test/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceITCaseTest.java#L1078-L1116) — asserts `NumberFormatException: For input string: "30s"` at query time on unmodified pre-fix code
  - Fix regression test (this PR): [`testHttpLookupJoinWithDurationRequestTimeout`](https://github.com/bharathgunapati/flink-connector-http/blob/FLINK-40072-request-timeout-duration/flink-connector-http/src/test/java/org/apache/flink/connector/http/table/lookup/HttpLookupTableSourceITCaseTest.java#L1078-L1116) — same DDL, join succeeds after the fix
- **Sink scope:** lookup reads via `ReadableConfig`; sink still reads from the `Properties` bag but parses with `TimeUtils.parseDuration(...)` — I left full sink config unification for a separate follow-up PR

Happy to adjust anything based on your feedback.

davidradl (commented):
Looks good . My AI had some nits:
1. Asymmetry in the sink path (moderate concern)
The lookup path reads via ReadableConfig.get(SOURCE_LOOKUP_REQUEST_TIMEOUT) — the proper typed API. The sink path reads via Properties.getProperty(SINK_HTTP_TIMEOUT_SECONDS) + TimeUtils.parseDuration(...).

// AbstractRequestSubmitter.java — current
String requestTimeout =
        properties.getProperty(HttpConnectorConfigConstants.SINK_HTTP_TIMEOUT_SECONDS);
this.httpRequestTimeout =
        requestTimeout == null
                ? DEFAULT_REQUEST_TIMEOUT
                : TimeUtils.parseDuration(requestTimeout);

This works, but it's a different mechanism than the lookup path. The sink presumably also has a ReadableConfig path (via HttpDynamicSinkConnectorOptions.SINK_REQUEST_TIMEOUT). If the sink's ReadableConfig is available here, using it directly (like the lookup does) would be more consistent. If Properties is the only option at this layer, a comment explaining why would help reviewers.

2. testLookupJoinWithUrlAsJoinKey_DISABLED — silent semantic change
// Before
'http.source.lookup.request.timeout' = '30'
// After  
'http.source.lookup.request.timeout' = '30s'

This is correct, but the test method name has _DISABLED — if this test isn't running in CI, the fix here is invisible in practice. Worth confirming the test actually executes (or noting it's a dormant fixture update).

3. requestTimeoutDefaultsToThirtySeconds test in BodyBasedRequestFactoryTest
RequestFactoryBase factory =
        new BodyBasedRequestFactory("GET", null, null, lookupConfig(new Configuration()));
assertThat(factory.httpRequestTimeout).isEqualTo(Duration.ofSeconds(30));

This relies on readableConfig != null (an empty Configuration is passed), so the default falls through readableConfig.get(SOURCE_LOOKUP_REQUEST_TIMEOUT), which returns Duration.ofSeconds(30). That's correct behavior, but the test exercises the readableConfig != null branch with a missing key — it might be worth a comment clarifying this is intentionally an empty config (not null).

4. HttpDynamicTableSinkFactoryTest#requestTimeoutOptionTest — weakened assertion
The old assertion:

assertThatThrownBy(...).isNotInstanceOf(ValidationException.class);

The new assertion:

try {
    tEnv.executeSql(...).await();
} catch (Exception e) {
    assertThat(e).isNotInstanceOf(ValidationException.class);
}

If the INSERT succeeds without throwing, the catch block is never entered and the assertThat never runs — which is fine, but it means the test passes vacuously in the success case. This is a minor testing style nit: the original was cleaner because it guaranteed an assertion always executed. The comment says "any other outcome is acceptable," which is accurate — but the approach is slightly harder to reason about. A plain assertThatCode(() -> ...).doesNotThrowAnyException() or assertThatCode(() -> ...).as("must not be a ValidationException").satisfies(...) would be cleaner.

5. httpRequestTimeout field is package-private in test assertions
// BatchRequestSubmitterTest
assertThat(submitter.httpRequestTimeout).isEqualTo(Duration.ofSeconds(45));

httpRequestTimeout is protected in the abstract class, which means test access from a different package only works because of test-class structure. This is fine for the test package but worth confirming the access modifier is intentional (vs. adding a getter or using a test subclass).

bharathgunapati:
Thanks for the review, @davidradl! Addressed the nits:

**Sink asymmetry** — added an inline comment in AbstractRequestSubmitter explaining the sink still reads from Properties today; full ReadableConfig wiring is planned as a follow-up PR.

**testLookupJoinWithUrlAsJoinKey_DISABLED** — noted that '30s' is a dormant fixture update (@Test is commented out); active coverage is testHttpLookupJoinWithDurationRequestTimeout.

**requestTimeoutDefaultsToThirtySeconds** — added a clarifying comment that empty Configuration exercises the option-default path.

**requestTimeoutOptionTest** — switched to catchThrowable with a null-safe check so the assertion always runs (including when INSERT succeeds).

**httpRequestTimeout** in tests — same-package access to protected is intentional; no getter added.

flink-connector-http/src/main/java/org/apache/flink/connector/http/table/lookup/RequestFactoryBase.java
davidradl: the AI is saying the points have been addressed and raises a nit — requestTimeoutOptionTest assertion
The catchThrowable approach in http/src/test/java/org/apache/flink/connector/http/table/sink/HttpDynamicTableSinkFactoryTest.java:127 still passes vacuously when INSERT succeeds (returns null):

Throwable insertFailure =
        catchThrowable(() -> tEnv.executeSql("INSERT INTO httpTimeout VALUES (1)").await());
assertThat(insertFailure instanceof ValidationException)
        .as("request.timeout must be a recognized option")
        .isFalse();

insertFailure instanceof ValidationException evaluates to false whether insertFailure is null (success) or any non-ValidationException throwable — so the assertion never actually validates the contract in the success case. The cleaner form that was raised in the earlier review round would be:

assertThatCode(() -> tEnv.executeSql("INSERT INTO httpTimeout VALUES (1)").await())
        .as("request.timeout must not raise ValidationException")
        .doesNotThrowAnyExceptionOfType(ValidationException.class);
// or if INSERT may legitimately fail for runtime reasons:
Throwable t = catchThrowable(...);
if (t != null) {
    assertThat(t).isNotInstanceOf(ValidationException.class);
}

The original assertThatThrownBy(...).isNotInstanceOf(ValidationException.class) assumed an exception always throws — replacing it with a null-safe check is the right idea, but the current form inverts the test logic in a way that makes the assertion always pass. This isn't a blocker (the intent is clear and the other tests cover the real contract), but it's worth a one-line fix before merge.

bharathgunapati: Updated requestTimeoutOptionTest to use explicit null-check + isNotInstanceOf(ValidationException) — only asserts when INSERT throws, which matches the test intent (runtime failures are acceptable, ValidationException is not).

bharathgunapati:
@davidradl Just a friendly  nudge to review this PR when you have a chance

PR #53, opened 2026-07-14
[FLINK-40149][connector-http] Rename asyncPolling to async-polling

bharathgunapati:

The HTTP lookup option for sync vs async polling is the only camelCase table
option in the connector (`asyncPolling`). All other options use kebab-case or
dotted keys (e.g. `lookup-method`, `http.source.lookup.request.timeout`).

This renames the canonical key to `async-polling` to align with Flink connector
naming conventions. The legacy `asyncPolling` key remains supported via
`withDeprecatedKeys` so existing DDL does not need to change.

No runtime behavior change: default remains `false`, and the option is already
read via `ReadableConfig.get(ASYNC_POLLING)`.

## Brief change log

- `HttpLookupConnectorOptions.ASYNC_POLLING`: canonical key `async-polling`,
  deprecated alias `asyncPolling`, description updated.
- Docs (en + zh): option table and examples updated to `async-polling`.
- Tests: parameterized factory and runtime tests cover both canonical and
  deprecated keys; ITCases updated to canonical key (mechanical).

## Compatibility note

- **Canonical key:** `async-polling`
- **Legacy key:** `asyncPolling` still works via `withDeprecatedKeys`
- **Default:** `false` (unchanged)
- **Migration:** optional — existing DDL using `asyncPolling` continues to work;
  new configurations should use `async-polling`

Release note is on [FLINK-40149](https://issues.apache.org/jira/browse/FLINK-40149).

## Verifying this change

CI green on fork (Flink 2.2.0, JDK 11/17/21):
https://github.com/bharathgunapati/flink-connector-http/actions/runs/29353591657

Automated tests:
- `HttpLookupTableSourceFactoryTest#shouldAcceptAsyncPollingAtFactory` — factory
  accepts both `async-polling` and `asyncPolling`
- `HttpLookupTableSourceTest#shouldCreateAsyncTableSourceWithAsyncPollingKey` —
  runtime async path for both keys

- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: **no**
- The serializers: **no**
- The runtime per-record code paths (performance sensitive): **no**

## Documentation

- Does this pull request introduce a new feature? **no** (consistency improvement)
- If yes, how is the feature documented? docs updated (en + zh)


bharathgunapati:
Hi @davidradl — would you be able to review this when you have a moment?

Small consistency improvement ([FLINK-40149](https://issues.apache.org/jira/browse/FLINK-40149)): renames the lookup option `asyncPolling` → `async-polling` to match Flink kebab-case conventions. The legacy key still works via `withDeprecatedKeys("asyncPolling")`, so no DDL migration is required.

**Scope:** option key + docs + tests only — no runtime logic change (default remains `false`, still read via `ReadableConfig`).

**Tests:** parameterized factory and runtime tests cover both `async-polling` and the deprecated `asyncPolling` alias.

**CI:** green on my fork (Flink 2.2.0, JDK 11/17/21) — https://github.com/bharathgunapati/flink-connector-http/actions/runs/29353591657

Thanks!

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -250,7 +250,7 @@ Note the options with the prefix _http_ are the HTTP connector specific options,
...
-| asyncPolling                                                           | optional | true/false - determines whether Async Polling should be used. Mechanism is based on Flink's Async I/O.                                                                                                                                                                                                                                                                                                 |
+| async-polling                                                          | optional | true/false - determines whether async polling should be used. Mechanism is based on Flink's Async I/O. Deprecated alias: `asyncPolling`.                                                                                                                                                                                                                                                                 |
```
davidradl: now we have release we cannot easily change a config name . We would need to deprecate but keep the old property and indicate people should use the new property name in preference. So we need to keep the docs for asyncPolling and still accept that property. We probably need some validation to reject configs with  both keys set

bharathgunapati: Agreed with the review feedback — pushed 0b4cb59:

- Docs list both keys: async-polling (preferred) and asyncPolling (deprecated, still accepted)
- Factory rejects DDL that sets both keys (ValidationException)
- asyncPolling still works at runtime via withDeprecatedKeys — no behavior change for existing DDL

davidradl (changes requested):
Looking good. My AI asked for some minor changes 

COMMENT 1
HttpLookupTableSourceFactory.java
Location: the 1-arg validateHttpLookupSourceOptions overload

The preserved 1-arg overload delegates to the 2-arg version with an empty map:

    protected void validateHttpLookupSourceOptions(ReadableConfig tableOptions) {
        validateHttpLookupSourceOptions(tableOptions, Collections.emptyMap());
    }

Because createDynamicTableSource now calls only the 2-arg version, a subclass
that overrides the 1-arg signature won't get the dual-key check applied. Is this
overload needed to preserve binary compatibility with any known subclass? If not,
I'd suggest removing the 1-arg method and keeping only the 2-arg signature — it
removes the override hazard and simplifies the code.

COMMENT 2
HttpLookupConnectorOptions.java
Location: DEPRECATED_ASYNC_POLLING_KEY constant

Minor: worth adding a comment near the DEPRECATED_ASYNC_POLLING_KEY constant (or
in its Javadoc) noting that Flink's withDeprecatedKeys emits a WARN log at
runtime when the deprecated key is used. This way future maintainers know the
runtime warning is intentional and expected behaviour, not a bug.

COMMENT 3
docs/content/docs/connectors/table/http.md
      docs/content.zh/docs/connectors/table/http.md
Location: asyncPolling row in the options table

Nit: the deprecation description currently reads "Do not set both keys in the
same DDL". Suggest rephrasing to be more direct and consistent with how Flink
typically words these:

  Deprecated. Use `async-polling` instead. Setting both keys in the same DDL
  is a validation error.

bharathgunapati:
@davidradl — thanks for the follow-up review. I’ve addressed the three comments in 37a4f22:

Removed the 1-arg validateHttpLookupSourceOptions overload — only the 2-arg signature remains, so there’s no subclass override hazard.
Documented the deprecated-key WARN — Javadoc on DEPRECATED_ASYNC_POLLING_KEY notes that Flink’s withDeprecatedKeys emits a WARN at runtime and that this is intentional.
Tightened the docs wording (en + zh) — asyncPolling now reads: Deprecated. Use async-polling instead. Setting both keys in the same DDL is a validation error.
Would you mind taking another look when you have a moment? Thanks!

PR #54, opened 2026-08-01
[FLINK-40277][connector-http] Add HTTP sink status-code handling and retries

bharathgunapati:

The HTTP sink currently treats non-success responses as generic request failures.
This PR adds configurable sink status-code handling and client-level retry handling so sink behavior is closer to the lookup source status-code/retry model.

The sink can now handle HTTP status codes as successful, ignored, retryable, or fatal. Retryable responses are retried by the sink HTTP client up to the configured retry limit before the sink writer reports failure.

## Brief change log

- Add configurable sink status-code options for success, retry, and ignored HTTP responses.
- Add sink retry configuration and retry strategy options.
- Add status-code handling for success, ignored, retryable failure, and fatal failure.
- Retry retryable sink failures at the HTTP client layer until `http.sink.max-retries` is exhausted.
- Preserve legacy `http.sink.error.code.exclude` behavior by mapping it to ignored response codes when the new ignored-response option is not set.
- Add unified sink configuration used by both Table API and DataStream API paths.
- Update sink HTTP client APIs to receive a `SinkHttpClientContext`.

## Compatibility note

- This PR changes `@PublicEvolving` APIs, including sink request/client-related types.
- `HttpSinkRequestEntry` construction has changed to use method and payload bytes.
- `SinkHttpClientBuilder` now builds clients from `SinkHttpClientContext`.
- Existing legacy `http.sink.error.code.exclude` configurations continue to work.
- `HttpSinkWriterStateSerializer` remains at version 1.
-  Fatal HTTP sink failures (network errors, non-retryable / fatal status codes, and exhausted retries) now fail the Flink job via completeExceptionally. Previously those paths called complete() after metrics/logging, so the job kept running and the records were effectively dropped. Jobs that used http.sink.error.code to mark failures (e.g. 5xx) and relied on continue-after-error behaviour will need to remap those codes or wait for continue-on-error / error-DLQ support in [FLINK-40699](https://issues.apache.org/jira/browse/FLINK-40699).

## Verifying this change

Automated tests include:

- `HttpSinkResponseClassifierTest`
- `HttpSinkWriterTest`
- `HttpSinkWriterStateSerializerTest`
- `HttpSinkConfigSerializationTest`
- `HttpDynamicTableSinkFactoryTest`
- `JavaNetSinkHttpClientTest`
- `HttpSinkConnectionTest`

Ran locally with JDK 17:

`mvn -pl flink-connector-http -DskipTests compile`

Targeted test command:

`mvn -pl flink-connector-http -DskipITs -Dskip.archunit=true -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dtest=HttpSinkWriterTest,HttpSinkResponseClassifierTest,HttpSinkConfigSerializationTest,HttpSinkWriterStateSerializerTest,HttpDynamicTableSinkFactoryTest,HttpSinkConnectionTest,JavaNetSinkHttpClientTest test`

- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: **yes**
- The serializers: **no**, `HttpSinkWriterStateSerializer` remains version 1
- The runtime per-record code paths (performance sensitive): **yes**, sink response handling now applies status-code handling and retries retryable responses

## Documentation

- Does this pull request introduce a new feature? **yes**
- If yes, how is the feature documented? Follow-up documentation work is tracked under the FLINK-40275 sink umbrella.

## Follow-up

FYI: this PR is scoped to configurable sink status-code handling and client-level retry handling. Broader sink refactorings are split across follow-up PRs in my fork: https://github.com/bharathgunapati/flink-connector-http/pulls.


bharathgunapati:
Hi @davidradl , I've captured the overall sink refactoring approach and subtasks in FLINK-40275, and this PR covers the first task in the list. Could you please take a look when you get a chance?

davidradl:
I suggest not using the term classification. I think you mean status codes.

bharathgunapati:
Thanks @davidradl, that makes sense. I’ve updated the PR/user-facing wording to use “status-code handling.”

I kept `HttpSinkResponseClassifier` internally because it maps a raw HTTP response to the sink’s semantic outcome (`SUCCESS`, `IGNORED`, `RETRYABLE_FAILURE`, or `FATAL_FAILURE`), so “classifier” describes the implementation role rather than the feature name.

davidradl:
@bharathgunapati I hope to get to this tomorrow

davidradl:
Just put this into my AI and it came back with:


1. HttpSinkClientWithRetry does not retry on network exceptions
In HttpSinkClientWithRetry.send(), the retry loop is wired only via retryOnResult — it retries when hasRetryableRequests() returns true. However, if the Java HTTP client throws (e.g., IOException for a connection timeout), the CompletionStage completes exceptionally. Resilience4j's decorateCompletionStage with only retryOnResult will not retry on exceptions unless retryOnException is also configured.
Yet in HttpSinkResponseClassifier.classify(), null response → RETRYABLE_FAILURE — the intent is clearly that network failures should be retried. But that path is only reached when optResponse.orElse(null) is null inside prepareAttemptResult, which only runs if submitRequests succeeds. If the future itself fails, the path never reaches the classifier.
Suggested fix: Add .retryOnException(IOException.class::isInstance) (or similar) to the RetryConfig built in HttpSinkClientWithRetry, and add a test case for it.
2. Default SINK_MAX_RETRIES = 3 changes behaviour for existing deployments
The option has a hard default of 3 in HttpDynamicSinkConnectorOptions. Previously retries were not performed at all. Any existing deployment that doesn't set this option will silently start retrying on 500/503/504, potentially tripling the number of HTTP calls on degraded endpoints. This is a silent behaviour change for existing users.
This should either default to 0 (explicit opt-in), or be clearly called out in the PR description / NOTICE / migration guide, and the compatibility note in the PR body updated.
3. SinkRetryConfigProvider will throw at runtime if maxRetries == 0
return createBuilder().maxAttempts(sinkConfig.getMaxRetries() + 1).build();

When maxRetries = 0, maxAttempts = 1 — that's valid, Resilience4j will make exactly one attempt. But createBuilder() also unconditionally reads getRetryStrategy() and tries to resolve RetryStrategyType.fromCode(...). If strategy config is missing and retries are disabled, this still constructs and validates the full retry configuration. Consider short-circuiting when maxRetries == 0 to avoid spurious validation errors and unnecessary overhead.
4. IGNORED responses are added to successfulRequests in the accumulator
In ResponseAccumulator.add(), IGNORED entries flow into successfulRequests. They're excluded from fatalFailedRequests and failedRequests, so the retry accumulator never sees them — but then SinkHttpClientResponse.successfulRequests conflates "succeeded" with "ignored", which may mislead callers who inspect getSuccessfulRequests() for post-callback processing. Consider a dedicated ignoredRequests field on SinkHttpClientResponse, even if the writer just combines them. This would also align the response shape with the four-way HttpSinkResponseStatus enum.
5. legacyResponseChecker short-circuits before the new ignoredResponseCodes check
In HttpSinkResponseClassifier.classify():
if (ignoredResponseCodes.contains(response.statusCode())) {
    return HttpSinkResponseStatus.IGNORED;   // ✓ checked first
}
if (legacyResponseChecker != null) {         // ← then legacy takes over entirely

If a user is on the legacy path and sets sink.ignored-response-codes, the ignored check wins — which is the correct precedence. But if a user is on the legacy path and does NOT set ignored codes, the new responseChecker is never consulted. That's intentional for compatibility, but it means sink.success-codes and sink.retry-codes are silently ignored when legacy error-code properties are present. This should be documented or at least logged as a warning.
6. DefaultSinkHttpClientContext is not Serializable
HttpSinkInternal.DefaultSinkHttpClientContext implements SinkHttpClientContext but is not serializable. The context is created transiently in buildSinkHttpClient(), so this is safe today. But SinkHttpClientContext itself is @PublicEvolving and users implementing custom SinkHttpClientBuilder will receive it — if they ever try to capture it across a serialisation boundary, they'll get a NotSerializableException at runtime. Consider either marking SinkHttpClientContext with extends Serializable or adding a Javadoc warning.
7. HttpSinkConfig has a redundant getReadableConfig() method
HttpSinkConfig stores a Configuration field (named readableConfig) and adds a method getReadableConfig() returning ReadableConfig. Lombok's @Getter would already generate getReadableConfig() returning Configuration. The custom override returning the ReadableConfig supertype while the @Getter-generated field is Configuration will cause a compilation conflict. This should be verified — it may require suppressing the Lombok getter for that field specifically.
8. HttpSinkBuilder.resolveSinkConfig() will NPE if endpointUrl is null
HttpSinkConfigFactory.fromDataStream(endpointUrl, ...) is called eagerly in build() before the existing null/blank check on endpointUrl that was in HttpSinkInternal. That check now operates on sinkConfig.getUrl() inside HttpSinkInternal, so the validation still happens — but it happens after resolveSinkConfig() already ran. If endpointUrl is null, fromDataStream passes it into the config builder without validation. The Preconditions check should move to resolveSinkConfig() or HttpSinkBuilder.build().
9. Deprecated SinkHttpClientResponse constructor uses HttpRequest.getRequestEntries()
The new compatibility constructor flattens List<HttpRequest> via request.getRequestEntries(). HttpRequest previously had no requestEntries field; this PR adds it as a constructor parameter. However, old custom clients that construct HttpRequest using the old three-arg constructor will get a NoSuchMethodError at runtime — the deprecated constructor doesn't actually provide binary compatibility. This should be noted in the Javadoc or the field given a nullable default path.
10. The SINK_WRITER_THREAD_POOL_SIZE default changed from "4" (string) to 1
HttpSinkWriter previously defaulted to 4 threads (HTTP_SINK_WRITER_THREAD_POOL_SIZE = "4"). HttpDynamicSinkConnectorOptions.SINK_WRITER_THREAD_POOL_SIZE now has .defaultValue(1). This is a silent performance regression for existing users. Intentional?

bharathgunapati:
Thanks for running this through, @davidradl — really useful list. Pushed `b085416` addressing it. Point by point:

**1. Retry on network exceptions.** Exception retries are now explicitly scoped to `IOException` (unwrapping `CompletionException`) in `HttpSinkClientWithRetry`, matching `HttpClientWithRetry` on the lookup side. One correction to the analysis: Resilience4j's default exception predicate is `throwable -> true`, so exceptions were already being retried — the change makes the intent explicit and, more importantly, stops non-retryable errors (e.g. a user callback throwing) from re-submitting the whole batch. Tests added in `HttpSinkClientWithRetryTest` and `JavaNetSinkHttpClientTest`.

**2. Default `max-retries`.** Agreed, this was a silent behaviour change. `http.sink.max-retries` now defaults to `0`, so existing jobs keep send-once behaviour and retries are opt-in. The connection tests that relied on a retry now set it explicitly.

**3. `maxRetries == 0` still building a full retry config.** Fixed — `SinkRetryConfigProvider` now short-circuits to `RetryConfig.custom().maxAttempts(1)` and never resolves the retry strategy when retries are disabled. Since this is now the default path, there's a test covering an invalid strategy value with retries off.

**4. IGNORED lumped into `successfulRequests`.**  `SinkHttpClientResponse` already changes shape here (`fatalFailedRequests` is new), so doing the ignored split now means consumers adapt once instead of twice. Ignored responses land in `getIgnoredRequests()`; the writer still treats them as committed and non-retryable.

**5. Legacy error-code path ignoring the new options.** Kept legacy precedence for compatibility, but it is no longer silent: `HttpSinkResponseClassifier` logs a WARN naming both legacy properties and the `http.sink.success-codes` / `http.sink.retry-codes` options being ignored. Also called out on the option descriptions and in the table docs (en + zh).

**6. `DefaultSinkHttpClientContext` not `Serializable`.** Documented on `SinkHttpClientContext` that instances are created at writer open time and are not part of checkpoint or job-graph serialization, so they must not be retained or serialized. I did not add `Serializable`, since the object never crosses that boundary and it would widen the public API for a runtime-only type.

**7. Lombok `getReadableConfig()` conflict.** Not reproducible — Lombok's `@Getter` skips generation when a method of that name already exists. `javap` on `HttpSinkConfig` shows a single `getReadableConfig()` returning `ReadableConfig`. This is the same pattern as `HttpLookupConfig` on `main`.

**8. NPE if `endpointUrl` is null.** A null URL does not NPE: `HttpSinkConfigFactory` passes it through and `HttpSinkInternal` rejects it during `build()` with `IllegalArgumentException: The endpoint URL must be set when initializing HTTP Sink.` `HttpSinkBuilderTest#testNullUrl` already covers this and passes.

**9. Deprecated `SinkHttpClientResponse` constructor.** Expanded the `@deprecated` javadoc. For the record, the constructor itself *is* binary compatible — generics erase, so the descriptor is `(Ljava/util/List;Ljava/util/List;)V` in both versions. The real caveat is on the read side: `getSuccessfulRequests()` / `getFailedRequests()` keep the same erased descriptor but now return `HttpSinkRequestEntry` elements, so old compiled readers would link and then hit a `ClassCastException`. That is now stated explicitly in the javadoc.

**10. Writer thread pool default 4 → 1.** Keeping `1`, since that is what the docs have always documented. Now called out explicitly in the option description and in both datastream and table docs: older versions used 4 threads when the option was unset, and users can set the option to `4` to keep the previous throughput.

On the naming point from earlier: the user-facing wording is "status-code handling" throughout the PR and docs. I've kept `HttpSinkResponseClassifier` as an internal name because it maps a response to the sink's outcome (`SUCCESS` / `IGNORED` / `RETRYABLE_FAILURE` / `FATAL_FAILURE`) rather than describing a configurable option. Happy to rename if you'd still prefer it gone.

Full module test suite is green: 631 tests, 0 failures.

docs/content.zh/docs/connectors/datastream/http.md
```diff
@@ -61,22 +61,31 @@ These options are specified on the builder using the setProperty method.
...
 | sink.flush-buffer.size                                  | optional | The maximum size of a batch of entries that may be sent to the HTTP endpoint measured in bytes.                                                                                                                                                  |
 | sink.flush-buffer.timeout                               | optional | Threshold time in milliseconds for an element to be in a buffer before being flushed.                                                                                                                                                            |
-| flink.connector.http.sink.request-callback                | optional | Specify which `HttpPostRequestCallback` implementation to use. By default, it is set to `slf4j-logger` corresponding to `Slf4jHttpPostRequestCallback`.                                                                                          |
```
bharathgunapati, line 64: one docs follow-up worth calling out.

While documenting the new sink options, I found that the DataStream connector
page listed properties under a `flink.connector.http.*` prefix. The runtime
prefix has been `http.*` since the GetInData donation (`FLINK_CONNECTOR_HTTP =
"http."`), and the same page's code samples already used `http.sink.*`. So those
table/prose rows documented keys that were silently ignored if copied.

I've fixed this in `5d6bf70` / `95a71a1` for the DataStream docs (en + zh):

- corrected existing sink option keys to `http.*`
- added rows for the new status-code and retry-strategy options introduced by
  this PR
- kept `http.sink.max-retries` default `0`, and restored
  `http.sink.writer.thread-pool.size` default to `4` so unset jobs keep the
  previous runtime behaviour

Table docs already used the correct `http.*` prefix; they only needed the new
option rows.

docs/content.zh/docs/connectors/table/http.md
```diff
@@ -596,12 +596,21 @@ another format name.
 | http.sink.request-callback                | optional | Specify which `HttpPostRequestCallback` implementation to use. By default, it is set to `slf4j-logger` corresponding to `Slf4jHttpPostRequestCallback`.                                                                            |
 | http.sink.error.code                      | optional | List of HTTP status codes that should be treated as errors by HTTP Sink, separated with comma.                                                                                                                                     |
 | http.sink.error.code.exclude              | optional | List of HTTP status codes that should be excluded from the `http.sink.error.code` list, separated with comma.                                                                                                                      |
+| http.sink.success-codes                   | optional | Comma separated HTTP status codes considered as successful sink responses. Use [1-5]XX for groups and `!` for exclusions. The default is `2XX`. Ignored when the legacy `http.sink.error.code` properties are set. |
```
davidradl, line 599: I suggest we error if both legacy and new option are specified and only allow one or the other. If they are specifying the new options they likely want to use them but did not reliase to remove the old one.   same for the config below

docs/content/docs/connectors/datastream/http.md
```diff
@@ -61,22 +61,31 @@ These options are specified on the builder using the setProperty method.
...
 | sink.flush-buffer.size                                  | optional | The maximum size of a batch of entries that may be sent to the HTTP endpoint measured in bytes.                                                                                                                                                  |
 | sink.flush-buffer.timeout                               | optional | Threshold time in milliseconds for an element to be in a buffer before being flushed.                                                                                                                                                            |
-| flink.connector.http.sink.request-callback                | optional | Specify which `HttpPostRequestCallback` implementation to use. By default, it is set to `slf4j-logger` corresponding to `Slf4jHttpPostRequestCallback`.                                                                                          |
```
davidradl, line 64: I assume we should not removed existing config options from the code - as this would break migrations. In this case we should keep the docs - maybe with a deprecated indicator and suggestions on what should be used instead for each. We should not reprefix the legacy config options I think.

docs/content.zh/docs/connectors/datastream/http.md
```diff
@@ -61,22 +61,31 @@ These options are specified on the builder using the setProperty method.
...
-| flink.connector.http.sink.writer.thread-pool.size         | optional | Sets the size of pool thread for HTTP Sink request processing. Increasing this value would mean that more concurrent requests can be processed in the same time. If not specified, the default value of 1 thread will be used.                   |
-| flink.connector.http.sink.writer.request.mode             | optional | Sets the Http Sink request submission mode. Two modes are available: `single` and `batch`. Defaults to `batch` if not specified. |
-| flink.connector.http.sink.request.batch.size              | optional | Applicable only for `flink.connector.http.sink.writer.request.mode = batch`. Sets number of individual events/requests that will be submitted as one HTTP request by HTTP sink. The default value is 500 which is same as HTTP Sink `maxBatchSize` |
+| http.sink.request-callback                                | optional | Specify which `HttpPostRequestCallback` implementation to use. By default, it is set to `slf4j-logger` corresponding to `Slf4jHttpPostRequestCallback`.                                                                                          |
```
davidradl, line 64: I would assume we want continue on error so errors in the sink do not kill the job. I am not sure if we can push the content of the error somehow into the sink topic for errors. We could do this separately - maybe as a separate task in the umbrella Jira.

flink-connector-http/src/main/java/org/apache/flink/connector/http/config/HttpConnectorConfigConstants.java
```diff
@@ -138,6 +138,38 @@ public final class HttpConnectorConfigConstants {
     public static final String SINK_HTTP_WRITER_THREAD_POOL_SIZE =
             FLINK_CONNECTOR_HTTP + "sink.writer.thread-pool.size";
 
+    public static final String SINK_MAX_RETRIES = FLINK_CONNECTOR_HTTP + "sink.max-retries";
```
davidradl, line 141: nit: we could have a constant FLINK_CONNECTOR_HTTP_SINK as FLINK_CONNECTOR_HTTP + "sink"

flink-connector-http/src/main/java/org/apache/flink/connector/http/config/HttpConnectorConfigConstants.java
```diff
@@ -138,6 +138,38 @@ public final class HttpConnectorConfigConstants {
...
+
+    public static final String SINK_SUCCESS_CODES = FLINK_CONNECTOR_HTTP + "sink.success-codes";
```
davidradl, line 143: I wonder if we should not have sink at all in the options common to sink and lookup and only have sink or lookup if it is sink or lookup specific. We could use withDeprecatedKeys in the configuration. This would get rid of a lot of the duplicate config keys across sink and lookup. And we may be able to have common code to process the common options as far as possible. We could re-origanise the docs into a common section and then a section for sink and look up specifics.  
wdyt?

davidradl:
another AI review 
1. HttpSinkClientWithRetry creates a new Retry instance on every send() call
In HttpSinkClientWithRetry.send():

Retry retry = Retry.of("http-sink-connector", RetryConfig.<HttpSinkAttemptResult>from(retryConfig)…);

Retry.of() registers into a global Resilience4j registry under that name. If the name already exists (another sink, a test that ran before), it silently returns the first registered config, not the one you just built. This is a correctness risk in multi-sink or test JVMs. Fix: use a random name per instance, or build the Retry once in the constructor and reuse it (the requestEntriesToSubmit AtomicReference already handles per-call state).

2. Behaviour change: fatal failures now throw instead of silently completing
Before this PR, both the "no response" and "error status" branches called resultHandler.complete() (silent drop). After the PR, all three bad branches call resultHandler.completeExceptionally(...). This is a correctness improvement, but it's also a breaking behaviour change for any existing job that relied on silent-drop semantics. The compatibility note in the PR description mentions API changes but doesn't mention this operational change. Existing jobs using http.sink.error.code to mark 5xx as errors will now see those jobs fail unless the codes are mapped to an ignored set.

3. Verify ResponseAccumulator captures the final successful attempt
In HttpSinkClientWithRetry, responseAccumulator.add() is called inside the retryOnResult predicate. The thenApply callback only calls markRetriesExhausted when hasRetryableRequests() is true. If the final attempt succeeds (no retryable requests), the question is whether Resilience4j's retryOnResult callback still fires for that last attempt before stopping. If it does not, the last batch of successful requests is silently missing from the accumulated response. Worth adding an explicit test for a "succeeds on the Nth retry" path and asserting the full successful count.

4. ignoredResponseCodes is evaluated before the legacy checker, yet the log warning says legacy props disable the new options
In HttpSinkResponseClassifier.classify():

if (ignoredResponseCodes.contains(response.statusCode())) {
    return IGNORED;                   // evaluated even in legacy mode
}
if (legacyResponseChecker != null) { // legacy path
    …
}

When legacy properties are active, ignoredResponseCodes is populated from http.sink.error.code.exclude (via setLegacyIgnoredResponseCodesIfNeeded). So excluded codes are classified as IGNORED rather than as legacy-non-errors. This is semantically correct, but the warning log says the new success-codes / retry-codes options are fully ignored when legacy props are set — that is slightly misleading. The Javadoc or warning should clarify that ignoredResponseCodes continues to serve as the "excluded" set even in legacy mode.

5. HttpSinkConfig.getReadableConfig() vs Lombok @Getter duplicate
HttpSinkConfig has @Getter on the class (from Lombok) and also a hand-written getReadableConfig() with a narrowed ReadableConfig return type. The Lombok-generated getReadableConfig() would return Configuration, so the hand-written one hides it to return the interface. This is functional but confusing — the field is still accessible as Configuration via the Lombok getter on readableConfig. Consider either removing @Getter on readableConfig and relying solely on the hand-written method, or dropping the hand-written method if callers don't need the narrowed type.

6. HttpSinkConfigFactory DataStream path: manual property-to-config mapping is fragile
HttpSinkConfigFactory.fromDataStream() has ~12 individual getProperty()/set() blocks. Each new config key needs a corresponding block here. A table-driven approach (iterating over ConfigOption definitions and their HttpConnectorConfigConstants string counterparts) would be more maintainable and less likely to miss a new option in a follow-up PR.

docs/content.zh/docs/connectors/table/http.md
bharathgunapati: Agree. Sink creation now fails if legacy http.sink.error.code / http.sink.error.code.exclude are set together with the new http.sink.success-codes / http.sink.retry-codes, and if error.code.exclude is set together with http.sink.ignored-response-codes. Defaults on the new options do not count as set, so legacy-only jobs keep working.

docs/content/docs/connectors/datastream/http.md
bharathgunapati: Agree we shouldn’t drop working options. This was a docs bug: the DataStream page listed flink.connector.http.*, but since the GetInData donation the runtime prefix has been http. (FLINK_CONNECTOR_HTTP = "http."), and the same page’s code samples already used http.sink.*. Those documented keys were never read if copied.

We fixed the docs to match the runtime. Legacy options that do exist (http.sink.error.code / error.code.exclude) remain supported and documented; we didn’t remove or reprefix them.

docs/content.zh/docs/connectors/datastream/http.md
bharathgunapati: Makes sense. This PR treats fatal / exhausted failures as job failures (completeExceptionally) instead of silently completing.

Tracked continue-on-error and an optional error/DLQ sink as [FLINK-40699](https://issues.apache.org/jira/browse/FLINK-40699) under the FLINK-40275 umbrella.

flink-connector-http/src/main/java/org/apache/flink/connector/http/config/HttpConnectorConfigConstants.java
bharathgunapati: Added FLINK_CONNECTOR_HTTP_SINK = FLINK_CONNECTOR_HTTP + "sink." and routed the sink property keys through it. Used "sink." (with the trailing dot) so the constant matches the existing key prefix; runtime key strings are unchanged.

bharathgunapati:
> another AI review

1. Retry.of() / global registry — Not a correctness issue on Resilience4j 1.7.1. Retry.of(name, config) constructs a new RetryImpl; it does not look up or reuse a global registry entry. Registry behaviour is a different API (RetryRegistry). Creating a Retry per send() is a bit more allocation, but it lets retryOnResult close over that attempt’s accumulator. No change planned here.

2. Fatal failures now fail the job — Confirmed. Fatal / exhausted paths now call completeExceptionally instead of silently completing. That is intentional for this PR; jobs that previously counted and continued on http.sink.error.code can break. Called out in the PR compatibility note. Continue-on-error / optional error-DLQ is tracked as [FLINK-40699](https://issues.apache.org/jira/browse/FLINK-40699).

3. Accumulator on the final successful attempt — Already covered. Resilience4j invokes retryOnResult for every completed result, including the last one. HttpSinkClientWithRetryTest#testRetriesOnlyRetryableRequests asserts that after a retryable entry later succeeds, both entries are in the final successful list.

4. Legacy warning vs ignored codes — The warning no longer claims that new options are fully disabled when legacy props are set. It only notes that legacy error-code properties remain supported and points to the new options for new jobs. In legacy mode, exclude still maps into the ignored set and is applied first. Mixing legacy and new status-code options now fails at sink creation.

5. Lombok getReadableConfig() — Not a conflict. Lombok skips generating a getter when the method already exists; the hand-written method returns ReadableConfig. Same pattern as HttpLookupConfig.

6. Manual fromDataStream mapping — Agree it’s repetitive. Tracked typed DataStream configuration (and cleaner mapping) as [FLINK-40700](https://issues.apache.org/jira/browse/FLINK-40700) under FLINK-40275 rather than expanding this PR.

flink-connector-http/src/main/java/org/apache/flink/connector/http/config/HttpConnectorConfigConstants.java
bharathgunapati: Agree. We can follow the Kafka table pattern: shared connection-level options (and for Kafka, client settings under properties.*), with scan.* / sink.* only where behaviour differs.

For HTTP that would mean sharing truly common connection settings (and parsing/docs), while keeping sink- and lookup-specific keys where values may differ — e.g. a sink table with http.sink.request.timeout = 30s and a lookup table with http.source.lookup.request.timeout = 5s in the same job. Table options are per table, and DataStream sink/lookup each have their own config, so those don’t collide.

Tracked as [FLINK-40702](https://issues.apache.org/jira/browse/FLINK-40702) under FLINK-40275.

